PR #41, opened 2020-05-30
MACI factory

### LINKED ISSUE #52: Create MACI factory

xuhcc:
The end goal is to make the process of per-round MACI deployment trustless.

Changes:

- Downgraded to solidity 0.5 to make our contracts compatible with MACI contracts.
- Added new MACI factory contract.
  - Separate contract is needed, otherwise the size of `FundingRoundFactory` contract exceeds EIP170 limit.
  - I made modifications to MACI contracts to gain access to several structs and reduce gas costs: https://github.com/xuhcc/maci/commit/5202f3a2cc12863781c39453a1a3a654c6db7d6f (I'll probably submit these changes to upstream if this PR passes review).
  - Some MACI parameters are hardcoded, but some are being passed by deployer. I didn't put much thought into it yet.
- Added `deployMaci()` method to `FundingRoundFactory` contract.
  - Just a proof-of-concept, I didn't touch other parts of the contract, so the old witness-related methods are still there.

Resolves #52


contracts/contracts/MACIFactory.sol
```diff
@@ -0,0 +1,83 @@
...
+import { QuadVoteTallyVerifier } from 'maci/contracts/sol/QuadVoteTallyVerifier.sol';
+
+contract MACIFactory is Ownable, MACIParameters {
+  // Constants
+  uint8 private constant MACI_STATE_TREE_DEPTH = 4;
+  uint8 private constant MACI_MESSAGE_TREE_DEPTH = 4;
+  uint8 private constant MACI_VOTE_OPTION_TREE_DEPTH = 4;
+  uint8 private constant MACI_TALLY_BATCH_SIZE = 4;
+  uint8 private constant MACI_MESSAGE_BATCH_SIZE = 4;
+  uint256 private constant MACI_MAX_USERS = 15; // 2 ** MACI_STATE_TREE_DEPTH - 1
```
proofoftom, line 19: Why not just use the constants you mention in this comment?

contracts/contracts/MACIFactory.sol
```diff
@@ -0,0 +1,83 @@
...
+  uint256 private constant MACI_MAX_MESSAGES = 15; // 2 ** MACI_MESSAGE_TREE_DEPTH - 1
+  uint256 private constant MACI_MAX_VOTE_OPTIONS = 16;
```
proofoftom, line 21: Am thinking 1024 is a max that we want to shoot for

contracts/contracts/MACIFactory.sol
proofoftom: Also, it appears that there's 5 leaves instead of two now (So 5**MACI_STATE_TREE_DEPTH-1). https://github.com/appliedzkp/maci/pull/107

contracts/contracts/MACIFactory.sol
```diff
@@ -0,0 +1,83 @@
...
+
+contract MACIFactory is Ownable, MACIParameters {
+  // Constants
+  uint8 private constant MACI_STATE_TREE_DEPTH = 4;
+  uint8 private constant MACI_MESSAGE_TREE_DEPTH = 4;
+  uint8 private constant MACI_VOTE_OPTION_TREE_DEPTH = 4;
+  uint8 private constant MACI_TALLY_BATCH_SIZE = 4;
+  uint8 private constant MACI_MESSAGE_BATCH_SIZE = 4;
+  uint256 private constant MACI_MAX_USERS = 15; // 2 ** MACI_STATE_TREE_DEPTH - 1
+  uint256 private constant MACI_MAX_MESSAGES = 15; // 2 ** MACI_MESSAGE_TREE_DEPTH - 1
```
proofoftom, line 20: Same as above

contracts/contracts/MACIFactory.sol
xuhcc: >Why not just use the constants you mention in this comment?

Doing calculation onchain would increase gas usage. But maybe solc does the needed optimizations, I didn't test it.

>Also, it appears that there's 5 leaves instead of two now (So 5**MACI_STATE_TREE_DEPTH-1)

I'll update my fork.

xuhcc: >But maybe solc does the needed optimizations, I didn't test it.

That turned out to be true. There's no difference in gas usage if constant is defined as arithmetic experssion.

>I'll update my fork.

The size of MACI contracts increased significantly and unfortunately the size of the MACI factory contract now exceeds EIP170 limit (25133 > 24576). Looking for solution...

xuhcc: >The size of MACI contracts increased significantly and unfortunately the size of the MACI factory contract now exceeds EIP170 limit (25133 > 24576).

Solved by changing the `runs` parameter of solidity optimizer.

PR #43, opened 2020-06-06
Contracts: Allow owner to change MACI parameters

### LINKED ISSUE #42: MACI factory: setting MACI parameters
- Only owner can change MACI parameters.
- Parameters are expected to change rarely, so I think there's no reason to pass them to `deployMaci()` every time. A separate setter is preferred.
- We probably don't want individual setters for each parameter, but a single method to set them all at once.

xuhcc:
All MACI parameters are configurable now.

Owner should call `setMaciParameters()` method on the `FundingRoundFactory` contract, which in turn will call  `MACIFactory.setMaciParameters()`. This is because `MACIFactory` contract is owned by the `FundingRoundFactory`.

Alternatively, we can change the owner of `MACIFactory` and allow Owner to call `MACIFactory.setMaciParameters()` directly, but I chose current implementation because it creates a single entry point for all Owner's calls.

Resolves #42

### DIFF
--- a/contracts/contracts/FundingRoundFactory.sol
+++ b/contracts/contracts/FundingRoundFactory.sol
@@ -18,9 +18,6 @@ import './FundingRound.sol';
 contract FundingRoundFactory is Ownable, MACIPubKey {
   using SafeERC20 for IERC20;
 
-  // Constants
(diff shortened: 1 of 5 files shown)


auryn-macmillan:
This looks good to me.
@proofoftom and @adamstallard could you please take a look at this.

PR #48, opened 2020-06-09
Misc: Add license file

xuhcc:
How about GPL? :wink:


auryn-macmillan:
Yeah, I'm happy with GPL.
Merging.

PR #61, opened 2020-06-26
Dapp: Display list of recipients

### LINKED ISSUE #29: Display essential recipient info
Can likely hard-code into the UI (as opposed to pulling from distributed storage or from on-chain)

- name
- description
- url
- etc.

### LINKED ISSUE #53: Script for test round deployment
Create a buidler script that will do the following:

- Deploy ERC20 token contract.
- Set native token for funding round factory.
- Set coordinator.
- Contribute tokens to the matching pool.
- Register dummy recipients.
- Deploy funding round.
- Deploy MACI.

xuhcc:
Resolves #53
Partially implements #29

### DIFF
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -1,0 +1,23 @@
+# clr.fund contracts
+
+## Scripts
+
+### Deploy factory contract
+
+```
+yarn buidler run --network localhost scripts/deploy.ts
+```
+
+### Deploy test round
+
+This includes:
+
+- Deploying ERC20 token contract.
+- Adding tokens to the matching pool.
+- Adding recipients.
+- Deploying funding round.
+- Deploying MACI instance.
+
+```
+yarn buidler run --network localhost scripts/deployTestRound.ts
+```
--- a/contracts/package.json
+++ b/contracts/package.json
@@ -7,6 +7,7 @@
     "node": "buidler node --port 18545",
     "deploy:local": "yarn clean && yarn build && buidler run --network localhost scripts/deploy.ts",
     "deploy:goerli": "yarn clean && yarn build && buidler run scripts/goerli.ts",
+    "deployTestRound:local": "yarn buidler run --network localhost scripts/deployTestRound.ts",
     "test": "buidler typechain && buidler test",
     "clean": "rm -rf cache && rm -rf build"
   },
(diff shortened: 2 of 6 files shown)


contracts/package.json
```diff
@@ -7,6 +7,7 @@
     "node": "buidler node --port 18545",
     "deploy:local": "yarn clean && yarn build && buidler run --network localhost scripts/deploy.ts",
     "deploy:goerli": "yarn clean && yarn build && buidler run scripts/goerli.ts",
+    "deployTestRound:local": "yarn buidler run --network localhost scripts/deployTestRound.ts",
```
proofoftom, line 10: I don't think yarn is needed in this command but it probably doesn't hurt either.

contracts/README.md
```diff
@@ -0,0 +1,23 @@
...
+This includes:
+
+- Deploying ERC20 token contract.
+- Adding tokens to the matching pool.
+- Adding recipients.
+- Deploying funding round.
+- Deploying MACI instance.
+
+```
+yarn buidler run --network localhost scripts/deployTestRound.ts
```
proofoftom, line 22: Same as above: `yarn deployTestRound:local`

contracts/README.md
```diff
@@ -0,0 +1,23 @@
+# clr.fund contracts
+
+## Scripts
+
+### Deploy factory contract
+
+```
+yarn buidler run --network localhost scripts/deploy.ts
```
proofoftom, line 8: Couldn't this simply be `yarn deploy:local`?

package.json
```diff
@@ -23,7 +23,7 @@
     "test": "yarn workspaces run test",
     "test:contracts": "yarn workspace @clrfund/contracts run test",
     "test:web": "yarn workspace @clrfund/vue-app run test",
-    "deploy:local": "yarn workspace @clrfund/contracts run deploy:local",
+    "deploy:local": "yarn workspace @clrfund/contracts run deploy:local && yarn workspace @clrfund/contracts run deployTestRound:local",
```
proofoftom, line 26: I wonder if we'll always want a test round deployed, or rather if there's a case where we might not want to have one deployed for testing? If the latter is true in any case we might want to separate this into another script. Not a big deal though.

proofoftom (approved):

package.json
xuhcc: I think we'll want the test round most of the time.

We may want to see how frontend responds to the situation when no round is deployed, but in that case we can simply run:

```
yarn workspace @clrfund/contracts run deploy:local
yarn start:web
```

instead of 

```
yarn start:dev
```

Issue #60, opened 2020-06-26
Style adjustments for contracts package


xuhcc:
- Rearrange the declaration of variables to be uints, bools, etc. then events; rearrange functions where it makes sense.
- Consider variable packing for saving on gas in Solidity with the uints
more information here: https://fravoll.github.io/solidity-patterns/tight_variable_packing.html
- Configure eslint to check test files.

xuhcc:
@maxgrok I've installed eslint in #64, but I didn't configure it and didn't fix any lint errors, so this part of the issue is not done yet. I can take care of it if you want.

PR #65, opened 2020-06-29
Styling Adjustments for Contracts Package -- Issue #60

maxgrok:
I rearranged the uints, bools, etc in order. in both the **FundingRound.sol** and the **FundingRoundFactory.sol** for styling.

### DIFF
--- a/contracts/contracts/FundingRound.sol
+++ b/contracts/contracts/FundingRound.sol
@@ -14,19 +14,21 @@ import './FundingRoundFactory.sol';
 contract FundingRound is Ownable, MACIPubKey {
   using SafeERC20 for IERC20;
 
-  // ERC20 token being used
-  IERC20 public nativeToken;
+  uint256 public counter;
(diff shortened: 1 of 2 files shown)


contracts/contracts/FundingRound.sol
```diff
@@ -14,19 +14,21 @@ import './FundingRoundFactory.sol';
 contract FundingRound is Ownable, MACIPubKey {
   using SafeERC20 for IERC20;
 
-  // ERC20 token being used
-  IERC20 public nativeToken;
+  uint256 public counter;
```
xuhcc, line 17: Perhaps we can also change `counter` to something more specific like `contributorCount`?

maxgrok: Sure!

xuhcc: Please check: there are two more places where this variable is used

PR #68, opened 2020-06-30
Chore/init nuxt app

dekanbro:


xuhcc:
@dekanbro Can you explain the purpose of this pull request? Since I'm also working on the frontend stuff, we have to coordinate in some way. 

I don't really like the idea of rewriting everything from scratch. I also think that the existing tooling is better than the one you are proposing here (in particular, typescript would better suit our project, because MACI modules are written in typescript).

proofoftom:
@xuhcc sorry for the lack of transparency as the decision happened in a different channel I think, am happy to invite you to the Raid Guild Discord channel as well. We've discussed and intend to bring over all of your work. The Nuxt decision was mine as I prefer its structure, magic, and support by the Vue community. I'm happy to implement typescript if you're familiar, I don't believe the two setups to be mutually exclusive, the reasoning for not including it was that neither @dekanbro or myself are and it presented more of a hurdle than feature.

xuhcc:
@proofoftom 

>We've discussed and intend to bring over all of your work.

Could you or @dekanbro rebase this branch onto `xuhcc:show-recipients` (#61)?

>The Nuxt decision was mine as I prefer its structure, magic, and support by the Vue community.

OK, I'm not very familiar with Vue ecosystem so let's use Nuxt if you think it is better.

>I'm happy to implement typescript if you're familiar, I don't believe the two setups to be mutually exclusive, the reasoning for not including typescript was that neither @dekanbro or myself are familiar with it and it presented more of a hurdle than feature.

Yeah, we already use it in the `contracts` app and in the MACI repo, and I think that type checks would be useful for frontend development as well. I also think that we should use the same tooling for backend and frontend to make collaboration easier, so I suggest replacing web3 with ethers if possible.

>am happy to invite you to the Raid Guild Discord channel as well

I'd like to keep the discussuions in our channels (although maybe I'll join Raid Guild someday :wink:).

xuhcc:
@dekanbro It seems that you did a merge instead of rebase, and for some reason GitHub now shows the diff incorrectly and it's hard to review. Please do a rebase:

```
git checkout chore/init-nuxt-app
git rebase master
```

In the end you should get clean linear history; then you can just force-push it. I just did it myself here: https://github.com/clrfund/monorepo/commits/nuxt

dekanbro:
hmm i did do that. any other ideas?

xuhcc:
Maybe we're using different versions of git (I'm on 2.27.0).
I think we can leave it as is and probably squash commits later when PR gets accepted. The issue with GitHub showing worng commits is well-known, here's the solution that could work: https://stackoverflow.com/a/46782679/1868395

xuhcc:
On the subject of PR:
1. I still think that typescript is a must-have. I recommend switching to it right now, because if we don't do that, more work may be required as the codebase grows. I'm familiar with typescript and can help resolve any issues if they will arise (or I can just install it myself).
2. I recommend replacing web3 with ethers because we already use it in `contracts` and as far as I know ethers is a good modern library and ethereum community is slowly migrating to it.
3. Why did you remove the `yarn.lock` file? Currently we have single lockfile for both workspaces, and it seems to work fine. I suggest either restoring it in the root dir or creating a separate `yarn.lock` in the `contracts` workspace.
4. Let's move  `vue-app/.editorconfig` to the project's root.
5. When I'm trying to run `yarn install`, I get this error `error nanoid@3.1.10: The engine "node" is incompatible with this module. Expected version "^10 || ^12 || >=13.7". Got "11.15.0"`. We need to use node v11 because MACI modules require it.

_(that's a preliminary review, I didn't yet actually run the code because of 5.)_

proofoftom:
@xuhcc I followed your suggestion but somehow totally botched this, but it looks like you figured it out in https://github.com/clrfund/monorepo/commits/nuxt, why don't we just merge that?

xuhcc:
@proofoftom no worries, I fixed it :)

xuhcc:
3: yarn keeps creating the root `yarn.lock` when I run `yarn install`. It seems to be designed to work with single lockfile so we need to restore it.
5: `nanoid "^3.1.10"` requirement comes from the package called @nuxt/telemetry. We can probably use https://www.npmjs.com/package/nuxt-optout to replace `@nuxt/telemetry`.

proofoftom:
> 3: yarn keeps creating the root `yarn.lock` when I run `yarn install`. It seems to be designed to work with single lockfile so we need to restore it.
> 5: `nanoid "^3.1.10"` requirement comes from the package called @nuxt/telemetry. We can probably use https://www.npmjs.com/package/nuxt-optout to replace `@nuxt/telemetry`.

Yup, these both sound like good solutions to me 👍

xuhcc:
@dekanbro Unfortunately `nanoid` dependency is still causing issues on Node v11. After some experimentation I found the solution, here is the pull request based on your branch: #73.

PR #69, opened 2020-07-01
Publish messages

### LINKED ISSUE #59: Funding Round contract: Add method for batch submission of messages
To improve contributor's experience, we need a method for batch submission of messages. I think it should be implemented in `FundingRound` contract because the size of MACI contract is already approaching limits.

```solidity
function submitMessageBatch(Message[] memory _messages, PubKey[] memory _encPubKeys)
```

It seems that we also need to ensure that the first message is a key-change command.

https://github.com/appliedzkp/maci/blob/master/specs/01_introduction.md#high-level-process

>A later version of MACI will have to mitigate a vote-buying attack where a user may be bribed immediately when the voting period starts, such that their very first message is influenced by a briber. One solution to this is to have the contract to mandate that the first message that each participant sends is to change their key (see below).

xuhcc:
- Installed MACI packages ~~(currently from tarballs built on `declarations` branch, see https://github.com/appliedzkp/maci/issues/127)~~.
- Added method for batch submission of messages. 
- Also added tests for `MACI.publishMessage()`.

Closes #59.


proofoftom (approved):
Thanks for the super clean and readable code! This looks good to go :)

Issue #49, opened 2020-06-10
Factory Contract: Enable switching between different recipient curation mechanisms


xuhcc:
Add method for setting recipient registry proxy:

```solidity
function setRecipientRegistryProxy(address _registryProxy)
```

A registry proxy must implement the following interface (to be discussed):

```solidity
contract RecipientRegistryProxy {
    function isValidRecipient(address _recipient)
}
```

Then in our contracts we can use it like that:

```solidity
function addRecipient(address _recipient) {
    require(recipientRegistryProxy.isValidRecipient(_recipient));
}
```

auryn-macmillan:
I'd like to propose this contract as a template for our curation mechanism interface.
https://github.com/clrfund/registry/blob/master/contracts/Registry.sol

The curation mechanism must expose two functions, allowing our contracts to lookup up a registrant's address by index or index by address:
```
function getIndexByAddress(address _registrant) public returns (uint) {
    ...
  }
```
&
```
  function getAddressByIndex(uint _index) public returns (address) {
    ...
  }
```

auryn-macmillan:
I don't think the funding round or factory contracts need to have an `addRecipient` function, that can be handled in the registry contract.

Our funding round contract would use `getAddressByIndex()` inside of `claimFunds()`, to determine where to send funds allocated to a particular index by MACI.

xuhcc:
>I'd like to propose this contract as a template for our curation mechanism interface.

This interface seems simpler. But we need a method to remove recipients to free up slots and at the same time guarantee that recipient at the given index will not change in the middle of the round. How would you solve that?

xuhcc:
```
require(registrants.length <= 3125,"registry is full");
```
This limit is not a constant and can change from round to round. We need a way to ensure that the number of recipients in registry is less or equal to the maximum vote options limit in current MACI instance.

auryn-macmillan:
> This interface seems simpler. But we need a method to remove recipients to free up slots and at the same time guarantee that recipient at the given index will not change in the middle of the voting period. How would you solve that?

Assuming `MAX_VOTE_OPTIONS == 3125`, I don't know that I agree that we need a way to remove recipients from the list, at least not for round 0, and likely not for the following few rounds either.

That functionality can be added in an upgrade if/when it is necessary.

> This limit is not a constant and can change from round to round. We need a way to ensure that the number of recipients in registry is less or equal to the maximum vote options limit in current MACI instance.

[...]

Issue #19, opened 2020-04-16
Funding Round Contract: public fund-claiming


slgraham:
Anybody can claim funds on behalf of a given recipient

xuhcc:
We found that it's not currently possible to know and verify both the amount of voice credits spent on a vote option and the quadratic vote tally for that option.

https://hackmd.io/NKohKwKgQyeyeQLpLepKig

xuhcc:
We'll also need a way to verify the sum of tally results across all vote options. This number will be used to calculate the proportion of the matching pool that recipient can claim: `matchingPoolShare = quadVoteTallyResult / quadVoteTotalSum`.

PR #72, opened 2020-07-07
Contracts: Assign indices to registered recipients

xuhcc:
- Assigning index to each registered recipient with respect to max vote options in MACI (needed for #19).
- Introduced `RecipientRegistry` interface as a first step to pluggable recipient curation (#49).
- Decreasing of `voteOptionTreeDepth` parameter is now forbidden in order to prevent invalidation of vote option indices of registered recipients (see related discussion in #49).


proofoftom (approved):
LGTM, asked a clarifying question over in TG but is non-blocker so am merging :)

Issue #80, opened 2020-07-31
Migrate from ethers v4 to ethers v5


xuhcc:
https://docs.ethers.io/v5/migration/ethers-v4/

PR #86, opened 2020-08-01
Initial migration from ethers v4 to v5

PumpkingWok:
I have migrated `ethers` from `4.0.47` to `5.0.1` and also all dependencies needed.
I had to add a `try/catch` block when it calls `parseLog()`, it seems to retrieve an event outside of the contract.
Look to this issue, i retrieved the same error -> https://github.com/ethers-io/ethers.js/issues/733

Fixes: #80


vue-app/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "core-js": "^3.6.4",
-    "ethers": "^4.0.47",
+    "ethers": "^5.0.1",
```
xuhcc, line 15: Please check the `vue-app/src/views/Home.vue`, this file needs to be updated for ethers v5

contracts/tests/utils.ts
```diff
@@ -1,6 +1,6 @@
 import { Contract } from 'ethers';
-import { TransactionResponse } from 'ethers/providers';
-import { BigNumber } from 'ethers/utils/bignumber';
+import { TransactionResponse } from '@ethersproject/abstract-provider';
+import { BigNumber } from 'ethers';
```
xuhcc, line 3: This import could be merged into one line with a previous import of `Contract`.

contracts/tests/utils.ts
```diff
@@ -23,9 +23,15 @@ export async function getEventArg(
...
   const receipt = await transaction.wait();
   for (const log of receipt.logs || []) {
-    const event = contract.interface.parseLog(log);
+    let event;
+    try {
+      event = contract.interface.parseLog(log);
+
+    } catch (e) {
+      console.log("event not from this contract");
+    }
```
xuhcc, line 32: I think it would be cleaner if we just filter the logs by the contract address:
```js
if (log.address != contract.address) {
  continue
}
```

contracts/tests/utils.ts
PumpkingWok: sure, i will fix it

contracts/tests/utils.ts
PumpkingWok: yes, it is clearer. i will fix it.

vue-app/package.json
PumpkingWok: I'm not sure that i understood properly. Do you mean as `import` in `Home.vue` ? Thanks

xuhcc: `event.values` -> `event.args`, maybe something else.

Try to run the frontend locally: `yarn start:node` and then `yarn start:dev` in another termimal window.

PumpkingWok: I checked and `event.values` to `event.args` seems the only edit to do. thanks

xuhcc:
@PumpkingWok This PR is marked as `[WIP]`, do you plan to add something else here? Otherwise I would merge it.

PumpkingWok:
Hi @xuhcc,
Thanks for the fast response, no in theory it should be ok.

Issue #57, opened 2020-06-17
Factory contract: Add recipients with metadata


slgraham:
For Round 0, recipients will be added to the round along with their metadata. The address will be stored in the contract and the metadata will be emitted with an event.

- [FundingRoundFactory.addRecipients()](https://github.com/clrfund/monorepo/blob/8f4977a2713e787047141a628201443217e41894/contracts/contracts/FundingRoundFactory.sol#L68) emits an event that includes whatever metadata we want (address, name, URL, description, profile image, everest id?, other stuff?)

xuhcc:
In the beginning we will need project name, description and an IPFS hash of an image:

![Screenshot_2020-08-04 Figma](https://user-images.githubusercontent.com/1660460/89265539-52e04500-d63d-11ea-8f2c-1cfbd838ec0c.png)

PR #88, opened 2020-08-04
Add recipients with metadata

PumpkingWok:
I left `_metadata` as string , i added `name`, `description`, and `ipfsHash` as parameters.

Fixes: #57

### DIFF
--- a/contracts/contracts/FundingRoundFactory.sol
+++ b/contracts/contracts/FundingRoundFactory.sol
@@ -29,14 +29,14 @@ contract FundingRoundFactory is Ownable, MACISharedObjs, IVerifiedUserRegistry, 
   FundingRound[] private rounds;
 
   mapping(address => bool) private users;
-  mapping(address => string) public recipients;
+  mapping(address => bool) private recipients;
   mapping(address => uint256) private recipientIndex;
 
   // Events
   event NewContribution(address indexed _sender, uint256 _amount);
   event UserAdded(address indexed _user);
   event UserRemoved(address indexed _user);
-  event RecipientAdded(address indexed _fundingAddress, string _name, uint256 _index);
+  event RecipientAdded(address indexed _fundingAddress, string _metadata, uint256 _index);
   event NewToken(address _token);
   event NewRound(address _round);
   event CoordinatorTransferred(address _newCoordinator);
@@ -100,22 +100,32 @@ contract FundingRoundFactory is Ownable, MACISharedObjs, IVerifiedUserRegistry, 
   /**
     * @dev Register recipient as eligible for funding allocation.
     * @param _fundingAddress The address that receives funds.
-    * @param _name The display name of the recipient.
+    * @param _metadata The metadata info of the recipient.
     */
-  function addRecipient(address _fundingAddress, string calldata _name)
+  function addRecipient(address _fundingAddress, string calldata _metadata)
     external
     onlyOwner
   {
     // TODO: verify address and get recipient info from the recipient registry
     require(_fundingAddress != address(0), 'Factory: Recipient address is zero');
-    require(bytes(_name).length != 0, 'Factory: Recipient name is empty string');
-    require(bytes(recipients[_fundingAddress]).length == 0, 'Factory: Recipient already registered');
+    require(bytes(_metadata).length != 0, 'Factory: Metadata info is empty string');
+    require(recipients[_fundingAddress] == false, 'Factory: Recipient already registered');
     // TODO: implement mechanism for replacing registrants
     require(recipientCount < maciFactory.maxVoteOptions(), 'Factory: Recipient limit reached');
     recipientCount += 1;
-    recipients[_fundingAddress] = _name;
+    recipients[_fundingAddress] = true;
     recipientIndex[_fundingAddress] = recipientCount;  // Starts with 1
-    emit RecipientAdded(_fundingAddress, _name, recipientCount);
+    emit RecipientAdded(_fundingAddress, _metadata, recipientCount);
+  }
+
+  function getRecipients(
+    address _recipient
+  )
+    external
+    view
+    returns (bool)
+  {
+    return recipients[_recipient];
   }
 
   function getRecipientIndex(
--- a/contracts/contracts/FundingRoundFactory.sol
+++ b/contracts/contracts/FundingRoundFactory.sol
@@ -118,16 +118,6 @@ contract FundingRoundFactory is Ownable, MACISharedObjs, IVerifiedUserRegistry, 
     emit RecipientAdded(_fundingAddress, _metadata, recipientCount);
   }
 
-  function getRecipients(
-    address _recipient
-  )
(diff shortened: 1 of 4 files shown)


auryn-macmillan:
I think we can probably roll `name`, `description`, and `ipfshash` into the `_metadata` string, no?

contracts/contracts/FundingRoundFactory.sol
```diff
@@ -100,22 +100,32 @@ contract FundingRoundFactory is Ownable, MACISharedObjs, IVerifiedUserRegistry,
...
+  }
+
+  function getRecipients(
+    address _recipient
+  )
+    external
+    view
+    returns (bool)
+  {
+    return recipients[_recipient];
```
xuhcc, line 128: We won't use this method, so it could be removed. At the frontend recipient list is built using the data from `RecipientAdded` events.

contracts/contracts/FundingRoundFactory.sol
```diff
@@ -100,22 +100,32 @@ contract FundingRoundFactory is Ownable, MACISharedObjs, IVerifiedUserRegistry,
...
     require(recipientCount < maciFactory.maxVoteOptions(), 'Factory: Recipient limit reached');
     recipientCount += 1;
-    recipients[_fundingAddress] = _name;
+    recipients[_fundingAddress] = true;
     recipientIndex[_fundingAddress] = recipientCount;  // Starts with 1
-    emit RecipientAdded(_fundingAddress, _name, recipientCount);
+    emit RecipientAdded(_fundingAddress, _metadata, recipientCount);
+  }
+
+  function getRecipients(
```
proofoftom, line 121: Since we're returning a singular recipient this should probably be name `getRecipient`

vue-app/src/views/Home.vue
```diff
@@ -41,9 +41,10 @@ async function getData() {
   recipientFilter.fromBlock = 0;
   const recipients = (await provider.getLogs(recipientFilter)).map(log => {
     const event = factory.interface.parseLog(log);
+    const metadata = JSON.parse(event.args["1"]);
     return {
       address: event.args["0"],
-      name: event.args["1"]
+      name: metadata["name"]
```
proofoftom, line 47: Shouldn't we add the description here as well?

adamstallard:
> I think we can probably roll `name`, `description`, and `ipfshash` into the `_metadata` string, no?

I don't know much about ipfs so I could be way off here, but if we roll together key, value pairs that could be naturally separated using existing ipfs rules, then we may have to create our own spec for separating them (e.g. ";" and ","). Unless such a spec exists for combining key value pairs in `_metadata`, I would avoid creating one.

proofoftom:
> I think we can probably roll `name`, `description`, and `ipfshash` into the `_metadata` string, no?

It is, this just more explicitly structures the stored data as metadata rather than just the `name`.

contracts/contracts/FundingRoundFactory.sol
PumpkingWok: Hi @xuhcc, i had to add this method for pass `allows owner to add recipient`, if i have to take `recipients` private

proofoftom:
> > I think we can probably roll `name`, `description`, and `ipfshash` into the `_metadata` string, no?
> 
> I don't know much about ipfs so I could be way off here, but if we roll together key, value pairs that could be naturally separated using existing ipfs rules, then we may have to create our own spec for separating them (e.g. ";" and ","). Unless such a spec exists for combining key value pairs in `_metadata`, I would avoid creating one.

I may be way off in understanding this, but we're just using JSON objects to have a noSQL-esque flexible key:value data structure
```
{
  name: "Commons Stack",
  description: "Solve the tragedy of the commons..."
}
```

PumpkingWok:
Hi @auryn-macmillan, thanks for the response. Maybe i explained it wrong, i mean that right now i have substituted `_name` in `addRecipient` function with `_metadata`, as String. For doing test and populate the UI i passed a JSON stringify with `name`, `description` and `ipfsHash` parameters

contracts/contracts/FundingRoundFactory.sol
xuhcc: You can also remove the corresponding check from the `allows owner to add recipient` test, it's not needed anymore.

vue-app/src/views/Home.vue
PumpkingWok: Right now i have used only `name` parameters for populate the UI, for leaving the same structure, but if you want i can add the `description` as well.

adamstallard:
> Hi @auryn-macmillan, thanks for the response. Maybe i explained it wrong, i mean that right now i have substituted `_name` in `addRecipient` function with `_metadata`, as String. For doing test and populate the UI i passed a JSON stringify with `name`, `description` and `ipfsHash` parameters

Ok I looked at the code.  `_name` has become `_metadata` which is json.  Then the json is serialized (as a String) so it can be passed to a smart contract. The smart contract never deserializes it, but we read it back again in our front-end and deserialize it. It makes sense.

vue-app/src/views/Home.vue
adamstallard: We're not showing descriptions yet in the front-end?  What about images?

PumpkingWok: Right now it shows only `name` and `address` but if you want i can add `description` and `ipfsHash` without problems. Should the ipfs hash point to an image ?

xuhcc: >Should the ipfs hash point to an image ?

Yes, it should be an image hash. See an example I provided here: https://github.com/clrfund/monorepo/issues/57#issuecomment-668431574

We can rename this parameter to `imageHash` to clarify that.

PumpkingWok: Hi @xuhcc, would you like that i rename `ipfsHash` in `imageHash` ? Also, do i have to add `description` when i populate the UI ?

xuhcc: >Hi @xuhcc, would you like that i rename ipfsHash in imageHash ?

Yeah, I would prefer `imageHash`.

>Also, do i have to add description when i populate the UI ?

It's not necessary.

We need to update home page to show descriptions and images according to this mockup: https://www.figma.com/proto/HotZCx3zNptdm0wCzXtZKf/clr.fund?node-id=275%3A188&viewport=342%2C185%2C0.14486059546470642&scaling=scale-down-width 

Initially I planned to do that myself in another PR, but you can do that here if you want to.

xuhcc:
@PumpkingWok Please resolve conflicts (rebase & force-push is OK).

vue-app/src/views/Home.vue
PumpkingWok: Ok, i changed `ipfsHash` to `imageHash` and i readjust `Home.vue` to populate using `_metadata`. I left the same structure using only `name` and not `description` and `imageHash`.

PR #95, opened 2020-08-11
Allow anyone to call claimFunds()

PumpkingWok:
Added new `address _recipient` parameter in `claimFunds()` and substituted `msg.sender` with it within this function to allow any address to reclaim it.

### DIFF
--- a/contracts/tests/round.ts
+++ b/contracts/tests/round.ts
@@ -557,7 +557,7 @@ describe('Funding Round', () => {
 
       await expect(fundingRoundAsContributor.claimFunds(...recipientClaimData))
         .to.emit(fundingRound, 'FundsClaimed')
-        .withArgs(contributor.address, expectedClaimableAmount);
+        .withArgs(recipient.address, expectedClaimableAmount);
       expect(await token.balanceOf(recipient.address))
         .to.equal(expectedClaimableAmount);
     });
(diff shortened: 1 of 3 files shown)


contracts/tests/round.ts
```diff
@@ -503,6 +503,7 @@ describe('Funding Round', () => {
     const totalVotes = 100; // Math.sqrt(totalSpent)
     const recipientIndex = 3;
     const recipientClaimData = [
+      recipient.address, // recipient
```
xuhcc, line 506: A similar change is needed in `contracts/scripts/claim.ts`

PumpkingWok: Done

contracts/tests/round.ts
```diff
@@ -549,6 +551,17 @@ describe('Funding Round', () => {
         .to.equal(expectedClaimableAmount);
     });
 
+    it('allows address different than recipient to claim allocated funds', async () => {
+      await token.transfer(fundingRound.address, matchingPoolSize);
+      await fundingRound.finalize(totalSpent, totalSpentSalt)
+
+      await expect(fundingRoundAsContributor.claimFunds(...recipientClaimData))
+        .to.emit(fundingRound, 'FundsClaimed')
+        .withArgs(contributor.address, expectedClaimableAmount);
```
xuhcc, line 560: It should emit event with recipient's address

PumpkingWok: Done

Issue #100, opened 2020-08-16
Load 3box profile of the current user


xuhcc:
Load and show profile image.

xuhcc:
Some dependencies of 3box package require node v12, so we need to migrate MACI to v12 first: https://github.com/appliedzkp/maci/issues/172

PR #127, opened 2020-09-25
Update dependencies to support node v12

xuhcc:
Required for https://github.com/clrfund/monorepo/issues/100#issuecomment-696797487

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "workspaces": { 
     "packages": ["contracts", "vue-app"],
     "nohoist": [
-      "**/@nomiclabs/**"
+      "**/@nomiclabs/**",
+      "**/maci-cli/**"
     ]
   },
   "scripts": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -24,5 +25,8 @@
     "lint": "yarn workspaces run lint",
     "lint:contracts": "yarn workspace @clrfund/contracts run lint",
     "lint:web": "yarn workspace @clrfund/vue-app run lint"
+  },
+  "resolutions": {
+    "ganache-core": "2.12.0-beta.0"
```
adamstallard, line 30: I'll trust that you need this, but not sure why.

adamstallard (approved):

package.json
xuhcc: Earlier versions of `ganache-core` had the `scrypt` package in its dependency tree which is not compatible with node v12.

PR #160, opened 2020-10-21
add about details

auryn-macmillan:
Added some about details and a link to the blog and forum.

### DIFF
--- a/vue-app/src/App.vue
+++ b/vue-app/src/App.vue
@@ -64,7 +64,7 @@ html {
 }
 
 a {
-  color: $highlight-color;
+  color: $text-color;
   text-decoration: none;
 }
 
--- a/vue-app/src/App.vue
+++ b/vue-app/src/App.vue
@@ -5,6 +5,8 @@
       <div id="nav-menu">
         <router-link to="/">Home</router-link>
         <router-link to="/about">About</router-link>
+        <a href="https://blog.clr.fund" target=_blank>Blog</a>
+        <a href="https://forum.clr.fund" target=_blank>Forum</a>
       </div>
     </div>
     <div id="content">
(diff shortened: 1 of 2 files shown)


vue-app/src/App.vue
```diff
@@ -62,7 +64,7 @@ html {
 }
 
 a {
-  color: $text-color;
+  color: $highlight-color;
```
xuhcc, line 67: Please don't change the base stylesheet.

vue-app/src/views/About.vue
```diff
@@ -1,15 +1,37 @@
 <template>
   <div class="about">
-    <h1 class="content-heading">About</h1>
-    <p>Clr.fund is a protocol for efficiently allocating funds to public goods that benefit the Ethereum Network according to the preferences of the Ethereum Community. Clr.fund strives for credible neutrality, decentralization, permissionlessness, trustlessness, and pseudonymity.</p>
+    <h1 class="content-heading title">About</h1>
+    <p class="content">Clr.fund is a protocol for efficiently allocating funds to public goods that benefit the Ethereum Network according to the preferences of the Ethereum Community. Clr.fund strives for credible neutrality, decentralization, permissionlessness, trustlessness, and pseudonymity.</p>
+
+    <h2 class="content-heading">Why does clr.fund exist?</h2>
```
xuhcc, line 6: `.content-heading` is the class for top-level heading inside `#content` block. It should not apply to other elements. It is also redundant.

xuhcc (changes requested):
I think this page should be styled similarly to project page. I can do that if you want

vue-app/src/views/About.vue
auryn-macmillan: > I think this page should be styled similarly to project page. I can do that if you want

Yeah, that would be great!

PR #197, opened 2020-11-19
Improve BrightID verification process & suggest wrapeth.com on xDai

### LINKED ISSUE #190: BrightID verification: Do not sponsor user before linking
In order to skip sponsoring step for users who were already sponsored, we need to change workflow a bit:

>Here's another possible order that avoids too many sponsorship transactions:
>1.  Have them sign something with their ethereum address to "log in"
>2. check /verifications
>  a. if they aren't linked, create a deep link or QR.
>  b. if they are linked, but not sponsored, create a sponsor transaction for them to sign
>  c. if they are linked and sponsored, show their verification status.

### LINKED ISSUE #193: Show link to exchange if user does not have sufficient funds
For example if native token is WXDAI, we should show a link to wrapeth or honeyswap.

xuhcc:
Resolves #190
Resolves #193

### DIFF
--- a/vue-app/src/api/bright-id.ts
+++ b/vue-app/src/api/bright-id.ts
@@ -48,6 +48,8 @@ export class BrightIdError extends Error {
   constructor(code?: number) {
     const message = code ? `BrightID error ${code}` : 'Unexpected error'
     super(message)
+    // https://github.com/Microsoft/TypeScript/issues/13965#issuecomment-388605613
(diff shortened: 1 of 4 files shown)


xuhcc:
The delay for unsponsored users is too big. We'll be waiting for BrightID node update which increases frequency of checking the Sponsor event.

xuhcc:
Replaced by #214

PR #204, opened 2020-11-28
Use bytes32 to identify recipients

### LINKED ISSUE #200: Funding round contract: Use recipient ID instead of address
Several projects can share the same recipient address, so we need to update our contracts to use recipient ID instead of address. This will also make clrfund compatible with a broader range of recipient registries.

`claimFunds()` method should accept `bytes _recipientId` instead of `address _recipient`. Similarly, `RecipientRegistry.getRecipientIndex()` method should accept recipient ID. It should also implement `getRecipientAddress()` method.

This is needed for #36

xuhcc:
Resolves #200


xuhcc:
This PR will likely be superseded by another PR where we will identify recipients by index.

PR #312, opened 2021-03-17
RecipientRegistry: Use timestamp instead of block numbers

### LINKED ISSUE #292: RecipientRegistry: Use timestamp instead of block numbers
- The timestamp to block number conversion in `FundingRound` contract is unreliable and needs to be removed.
- MACI 1.0 will still use timestamps to specify voting period.
- Optimism and Arbitrum rollups both support `block.timestamp`.

xuhcc:
Resolves #292

### DIFF
--- a/vue-app/src/api/recipient-registry-kleros.ts
+++ b/vue-app/src/api/recipient-registry-kleros.ts
@@ -65,8 +65,8 @@ function decodeRecipientAdded(event: Event, columns: TcrColumn[]): Project {
 
 export async function getProjects(
   registryAddress: string,
-  startBlock?: number,
-  endBlock?: number,
+  startTime?: number,
+  endTime?: number,
 ): Promise<Project[]> {
   const registry = new Contract(registryAddress, KlerosGTCRAdapter, provider)
   const tcrAddress = await registry.tcr()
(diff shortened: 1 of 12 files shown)


vue-app/src/api/recipient-registry-kleros.ts
```diff
@@ -79,19 +79,19 @@ export async function getProjects(
...
     }
     const removed = recipientRemovedEvents.find((event) => {
       return (event.args as any)._tcrItemId === project.id
     })
     if (removed) {
-      if (!startBlock || startBlock && removed.blockNumber <= startBlock) {
-        // Start block not specified
-        // or recipient had been removed before start block
+      const removedAt = (removed.args as any)._timestamp.toNumber()
+      if (!startTime || startTime && removedAt <= startTime) {
```
adamstallard, line 92: The second `startTime` is logically redundant. Do you think it's more clear or less clear than this simplification?
```suggestion
      if (!startTime || removedAt <= startTime) {
```

vue-app/src/api/recipient-registry-simple.ts
```diff
@@ -39,17 +39,19 @@ export async function getProjects(
...
     }
     const removed = recipientRemovedEvents.find((event) => {
       return (event.args as any)._recipientId === project.id
     })
     if (removed) {
-      if (!startBlock || startBlock && removed.blockNumber <= startBlock) {
-        // Start block not specified
-        // or recipient had been removed before start block
+      const removedAt = (removed.args as any)._timestamp.toNumber()
+      if (!startTime || startTime && removedAt <= startTime) {
```
adamstallard, line 52: Same comment here as above. It also makes me wonder if this can be made more DRY since we're seeing this code again.

```suggestion
      if (!startTime || removedAt <= startTime) {
```

adamstallard (approved):
Approved, but may benefit from refactoring.

vue-app/src/api/recipient-registry-kleros.ts
xuhcc: I agree, it should be simplified.

vue-app/src/api/recipient-registry-simple.ts
xuhcc: >It also makes me wonder if this can be made more DRY since we're seeing this code again.

Not worth the increased complexity at the moment.

PR #340, opened 2021-04-15
Move mnemonics to .env file and differentiate RPC URLs.

auryn-macmillan:
Quick PR to move mnemonic out of hardhat.config.ts and into the .env file, so there is less chance of someone accidentally pushing their mnemonic to github.

### DIFF
--- a/contracts/.env.example
+++ b/contracts/.env.example
@@ -1,3 +1,4 @@
 RINKEBY_JSONRPC_HTTP_URL=https://eth-rinkeby.alchemyapi.io/v2/ADD_API_KEY
 XDAI_JSONRPC_HTTP_URL=https://rpc.xdaichain.com
-MNEMONIC=
+WALLET_MNEMONIC=
+WALLET_PRIVATE_KEY=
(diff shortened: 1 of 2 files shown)


contracts/.env.example
```diff
@@ -1 +1,3 @@
-ETHEREUM_JSONRPC_HTTP_URL=https://eth-rinkeby.alchemyapi.io/v2/ADD_API_KEY
\ No newline at end of file
+RINKEBY_JSONRPC_HTTP_URL=https://eth-rinkeby.alchemyapi.io/v2/ADD_API_KEY
+XDAI_JSONRPC_HTTP_URL=https://rpc.xdaichain.com
+MNEMONIC=
```
xuhcc, line 3: Let's rename it to `WALLET_MNEMONIC`. This setting can be complemented later with `WALLET_PRIVATE_KEY`

xuhcc (changes requested):

contracts/.env.example
auryn-macmillan: Done

PR #347, opened 2021-04-23
Choosing recipient registry type for local deployment

### LINKED ISSUE #346: Refactor dev deploy scripts to deploy registries based on ENV variable
In the frontend app we have a convenient ENV variable so we know which user & recipient registries to interact with. e.g. https://github.com/clrfund/monorepo/blob/develop/vue-app/.env.example#L14

It'd be great to add flexibility to our Hardhat deploy scripts to deploy the corresponding registries based on the same / similar ENV variable. Currently it always deploys the simple user & recipient registry:
https://github.com/clrfund/monorepo/blob/develop/contracts/scripts/deploy.ts#L30

xuhcc:
Resolves #346

### DIFF
--- a/contracts/.env.example
+++ b/contracts/.env.example
@@ -1,4 +1,8 @@
+# Recipient registry type for local deployment: simple, optimistic
+RECIPIENT_REGISTRY_TYPE=simple
+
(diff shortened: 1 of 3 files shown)


samajammin:
Amazing! You're a hero 🏆

PR #358, opened 2021-05-19
Add recipient registry info to store

samajammin:
@xuhcc looking to get your take on this...

Currently we store [recipientRegistryAddress in the store](https://github.com/clrfund/monorepo/blob/develop/vue-app/src/store/index.ts#L53).

In our [Eth2 CLR fund work](https://github.com/ethereum/clrfund) I found myself [querying registry info](https://github.com/clrfund/monorepo/blob/develop/vue-app/src/api/recipient-registry-optimistic.ts#L19) in multiple different components. I think the ideal approach here is to fetch that data once & add it to the Vuex store.

Please let me know what you think! Appreciate any feedback.


vue-app/src/App.vue
```diff
@@ -74,11 +80,23 @@ export default class App extends Vue {
...
+      this.$store.dispatch(LOAD_RECIPIENT_REGISTRY_INFO)
+    }, 60 * 1000)
     setInterval(() => {
       this.$store.dispatch(LOAD_USER_INFO)
     }, 60 * 1000)
   }
 
+  // TODO commit() approach? Or dispatch() on interval?
+  async mounted() {
+    const roundAddress = this.$store.state.currentRoundAddress || await getCurrentRound()
```
xuhcc, line 93: `this.$store.state.currentRoundAddress` will be set in `Project` or `ProjectList` component. I think you can't count on its value being correct here.

vue-app/src/store/index.ts
```diff
@@ -15,14 +15,17 @@ import {
   getContributionAmount,
   isContributionWithdrawn,
 } from '@/api/contributions'
+import { recipientRegistryType } from '@/api/core'
 import { loginUser, logoutUser } from '@/api/gun'
+import { getRegistryInfo, RegistryInfo } from '@/api/recipient-registry-optimistic'
```
xuhcc, line 20: Other types of registries (such as Kleros adapter) can have registry info too, but of different type.
There should be a wrapper function in `api/projects.ts`

vue-app/src/App.vue
```diff
@@ -74,11 +80,23 @@ export default class App extends Vue {
...
+    setInterval(() => {
+      this.$store.dispatch(LOAD_RECIPIENT_REGISTRY_INFO)
+    }, 60 * 1000)
     setInterval(() => {
       this.$store.dispatch(LOAD_USER_INFO)
     }, 60 * 1000)
   }
 
+  // TODO commit() approach? Or dispatch() on interval?
+  async mounted() {
```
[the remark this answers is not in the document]
xuhcc: Actions are usually created to avoid code duplication, so using `LOAD_RECIPIENT_REGISTRY_INFO` seems appropriate here

vue-app/src/store/index.ts
samajammin: I checked out `api/projects.ts` but I'm only seeing wrapper functions around getting/registering projects. I'm not seeing anything around registry info in `api/recipient-registry-kleros` but please let me know if I'm missing something you'd like me to add here (or if there's a wrapper function you'd like me to create).

samajammin:
Thanks for the review @xuhcc - I made some updates after reading your comments & thinking through this a bit more.

Reasoning on this approach:
- Refactored `App.vue` to dispatch actions (per your comment)
- When `App.vue` mounts, we set a default current round, so we can dispatch an action to load recipient info (since we need the current round address to get the recipient registry address)
- Commit the recipient address in the `LOAD_RECIPIENT_REGISTRY_INFO` action (in case we don't already have that in the store)

Again, please let me know what you think here. We need this functionality in our fork so that we can conditionally display calls to action for projects to apply (if the round is using an optimistic recipient registry). I figure this is logic this core project could benefit from as well.

vue-app/src/App.vue
```diff
@@ -74,11 +74,20 @@ export default class App extends Vue {
...
+    }, 60 * 1000)
     setInterval(() => {
       this.$store.dispatch(LOAD_USER_INFO)
     }, 60 * 1000)
   }
 
+  async mounted() {
+    const roundAddress = await getCurrentRound()
+    await this.$store.dispatch(SELECT_ROUND, roundAddress)
+    this.$store.dispatch(LOAD_RECIPIENT_REGISTRY_INFO)
```
xuhcc, line 88: This is not correct because the round address can also come from route parameter in `ProjectList` component.

See my previous comment on this section of code

xuhcc: Construction of the app's state begins with a round address, because almost everything else depends on it.

There are two ways in which this variable can be set:
1. If user lands on the project list page, the `ProjectList` component gets round address from route parameters.
2. If route parameters does not contain round address, or if user lands on some other page, the app gets round address from the `FundingRoundFactory` contract.

(1) is the reason why the initialization does not currently take place in the `App.created()`. Doing initialization in `App` should be possible, but it would require more work.

Your proposed change simply creates a race for dispatching `SELECT_ROUND` between `App` and `ProjectList` components.

samajammin: Thanks for the explanation.

> 2. If route parameters does not contain round address, or if user lands on some other page, the app gets round address from the FundingRoundFactory contract.

Where does this logic take place? Does it happen on page load / app initialization? I see we dispatch `LOAD_ROUND_INFO` every 60 seconds in App.vue but for our purposes we need to fetch round info on any initial page load. I'm open to refactoring this to accomplish that without creating a potential clash vs. the route parameters of `ProjectList`.

samajammin:
Closing this out. Realizing this change is more specific to the use case of https://github.com/ethereum/clrfund where we intend to have the frontend app tailored to a single funding round (vs. having the ability to view multiple rounds).

Thanks again for the feedback!

PR #364, opened 2021-06-22
Proposal: use of Prettier in the dev flow

pettinarip:
I want to have your opinion and/or suggestions with the inclusion of Prettier in the dev flow.

The goal is to improve the flow by adding `prettier` so that we have a consistent format style across the app codebase and to make it easier to maintain with new developers that want to contribute to it.

### Implementation

I've installed `husky` to use the git hooks so that we can format everything before each commit is done and avoid pushing things with the wrong format.

I'm installing everything on the root level and not in each package. **Why?:**
1. Because we want to have the same format style in all the packages we have in the repo. No need to have a `.prettierrc` in each package.
2. Because we need `husky` to be in the same folder as the `.git` is in. Husky doesn't work super nice with workspaces. I tried to install it only in the `vue-app` folder but it didn't work.

* * *

**Bonus:**
Since we've installed `husky`, I'm also using the `pre-push` hook to run unit tests and format checks to ensure that the state of the code we are pushing is good.
I'm running only the vue tests for this (if you look into the pre-push script), not the `/contract` ones since those tests are too many and it takes a lot of time to run them. And since we are not doing that many changes in that project I think it does not deserve to do them on every push.


xuhcc:
>The goal is to improve the flow by adding prettier so that we have a consistent format style across the app codebase and to make it easier to maintain with new developers that want to contribute to it.

Not sure what are you talking about. The code analysis tools are already configured -- you can simply run `yarn lint`.

pettinarip:
Well, the `yarn lint` command is running behind the scenes just eslint, a linter. Prettier is a formatting tool so it has some differences over the linter:
- The linter is not going to format the entire code for you. It will mostly only going to warn or error you about the wrong styles you have. Some of the ESlint rules are not "fixable".
- Prettier support a lot of different languages, not only JS. So, you can format JSON, CSS, SCSS, HTML, etc.
- Depending on the IDE you are using (nowadays all of them are integrated with prettier), you can benefit from automatic format styling when you save your files, using the same format definitions you have defined in your project (reading the `.prettierrc` file). This is super handy for the developer experience.

So basically, you end up using Prettier for formatting the code and using ESlint to check code quality. More about this: https://prettier.io/docs/en/comparison.html

codeluggage:
Great addition!

This makes a big impact when the CI/CD pipeline has linting as one of its steps, and an errant comma or double quotes stops the entire build. 

In terms of speed, and even more so when the repo grows in size, Husky can be kept small and maintainable with https://github.com/lerna/lerna too. 

Haven't tried the newer Husky, nor the new git hook approach, so I look forward to giving this a spin. 

LGTM from my limited experience with newer Husky 👍

xuhcc:
I prefer tooling that doesn't make automatic changes to the code or otherwise interferes with my work. If you need this, just use `eslint --fix` which works pretty well (for example, if you want to enforce a new linter rule).
Similarly, if you want to use git hooks, just configure them in your local repo, I don't see any reason to force other contributors to install them.

pettinarip:
> tooling that doesn't make automatic changes to the code

This is something completely handled by the user, by the configuration he has on his IDE.

> or otherwise interferes with my work

This proposal is not going to automatically change your code. It will only format it when you commit it. If the user wants, he can turn on the "format on save" option in his IDE and he will have automatic format.

> If you need this, just use eslint --fix

Unfortunately, this doesn't work as well as prettier with styling rules. Some styling rules are not automatically fixed by eslint, so you end up manually fixing them which is something bad (for the dx).

> if you want to use git hooks, just configure them in your local repo

What we want to achieve with this is to have a **consistent** code format across the codebase. Consequently, the code is not going to have different code styles => will be more readable => will be more maintainable.

> I don't see any reason to force other contributors to install them

Nowadays, this is a pretty standard setup in modern web apps. I think most of them will thank you for using them. You will catch errors before the CI/CD build does it for you. You will save time for them and for the reviewers.

xuhcc:
The linter is already configured, and it helps with keeping coding style consistent. Anyone who wants automatic formatting should configure it in their local development environment because this is simply a matter of choice. It is not a "standard".

If you want to contribute to this project, I suggest starting with something substantial and don't mess with established tooling.

samajammin:
Adding my 2 gwei for whatever it's worth: I think it makes sense to enforce code formatting on collaborative projects (which eslint alone does not accomplish). I'm fairly indifferent on _how_ the formatting is enforced or what specific rules to use but as @pettinarip mentioned, it helps keep the codebase more readable & maintainable. 

Otherwise (at least from my experience with OSS projects), you end up having various contributors with different code formatting approaches & a bunch of noise in PRs (because people keep re-formatting the same code in different ways).

auryn-macmillan:
If this is something that will make the code-base easier to maintain, easier on new contributors, doesn't make more difficult for existing contributors, and doesn't introduce any meaningful risks, I don't see any good reason to oppose it.

xuhcc:
It won't make codebase easier to maintain. There are huge projects that don't use autoformatting at all and they are doing just fine. Auto-formatting is the most intrusive and annoying way to enforce code style and it actually discourages new contributors rather than helps them.

What is happening here is that some people (who haven't contributed to the project in the past) are trying to replace established tooling by one they're more familiar with, that's all.

@samajammin If prettier and husky is so popular among EF employees, they can use it in your clrfund fork. Any problem with that?

pettinarip:
This was just a proposal to read your thoughts about it. I don't want to replace anything. I took the time to do this PR because I thought it could add value to the project (which I think it does) and you made it clear what your position is. The discussion is closed, at least on my side.

PR #376, opened 2021-07-09
ETH2 Features

daodesigner:
removed most of the linter and spoofed the commits so git blames roughly line up to what they were before the linter was run. 
To review features look at: 
#373 - adds new assets
#374 - augments API and Store
#375 - implements new components based on API and Store, updates router, etc

The linter and any copy related to ETH2 specifically was ommitted *( please note it is bad form to lint other's work as it removes code attribution). @wackerow please comment here with more context on the mentioned branches, and retroactively create issues describing each. This is about as reasonable to break up as I could manage. Screenshots below: 
image
image
image
image


daodesigner:
IMO the main screen is a little busy rn and has a bad gaze sequence projection.

## Visual Attention Study: 
#### Visual heatmap
busy busy_VAS_Heatmap
#### Gaze sequence
busy busy_VAS_VisualSequence

xuhcc:
What's the point of this PR? The diff is still huge and it changes the UI completely.

Did you read what I have said on the forum?

daodesigner:
As you can see this pull request is still a work in progress, I'll let you know when we would like it to be reviewed. 

Moreover, it's an olive branch from our end, I agree with you (now) that the only way to do this cleanly is to reimplement everything, but I disagree this is the best use of everyone's time and am getting some BDFL vibes. I split the diff into three branches which we can alternatively merge separately (this branch is just for convenience and for me to verify the diff with their fork looks right).

1) https://github.com/clrfund/monorepo/tree/feature/assets (new assets and a big chunk of the diff) 
2) https://github.com/clrfund/monorepo/tree/feature/augment-api-and-store (does not change UX)
3) https://github.com/clrfund/monorepo/tree/feature/new-ux-components/ (implements new components and UX changes)

They should be merged in this order and we can handle each under a different PR (1) & (2) would allow us to re implement everything in (3). If you you have concrete criticism about the UI changes now is definitely the time to say so. Should I split this into three PRs then? The diff for each is under 100 files and I've personally read through most of each of them now. 

It's fine if you want to decline the PR, just drop the attitude, I obviously read what you said on the forum. Let's try to wrap this up productively and professionally, there's no need for anyone to get worked up.

xuhcc:
On the forum I said that we need to decide which features to backport. But I don't see any discussion on this topic, only some vague descriptions in PRs you mentioned.

- _new assets_ - it seems to be useless without new UI.
- _api and store to support richer UX_ - all these changes are probably useful for ETH2CLR project, but I don't see how they can be an improvement to clrfund.
- _Landing page wth nav, Cart UX upgrade, modal upgrade, IPFS upload, info pages, tooltips._ - I would agree that landing page, info pages and IPFS upload are nice features. Everything else seems useless.

>If you you have concrete criticism about the UI changes now is definitely the time to say so. 

There's nothing to criticize, because it's not a collection of small changes. It's completely redesigned UI. And as a whole, it looks strictly worse that the current UI.

>Moreover, it's an olive branch from our end, I agree with you (now) that the only way to do this cleanly is to reimplement everything, but I disagree this is the best use of everyone's time and am getting some BDFL vibes. 

Lol, what does is supposed to mean? I still have no idea why you're trying to merge all this stuff into clrfund. If you just don't like current design, you can switch to ethereum/clrfund fork and stop wasting my time and yours.

daodesigner:
sounds good. Apologies for wasting your time, will try to maintain your code attribution in the new repos. Just trying to meet a deadline and I respect your decision.

PR #379, opened 2021-08-16
JoinLanding data fetching bug

### LINKED ISSUE #227: JoinLanding data fetching bug
https://www.notion.so/efdn/JoinLanding-data-fetching-bug-76b44882759542b3835d9b66f0e320d9

Something wacky is going on with our data fetching & rendering for this view.

If I load this page directly, e.g. opening a link:
https://eth2clrfund.netlify.app/#/join

No warning is displayed:
image

But if I navigate to a different page, then return I see this warning about the registry being almost full:
image

Let's investigate how we fetch this data & see if there's some issue e.g. if there's a race condition we're hitting.

**Note** - replicating this bug will require specific conditions (the MACI caps being almost full). - this can be mocked by updating the frontend values: https://github.com/ethereum/clrfund/blob/develop/vue-app/src/api/round.ts#L158

corwintines:
Notion: https://efdn.notion.site/JoinLanding-data-fetching-bug-76b44882759542b3835d9b66f0e320d9
Fixes: https://github.com/ethereum/clrfund/issues/227

Changes:
- Converted the created lifecycle method into getters since there was no reason to really leave it as a lifecycle method. This makes up for any race condition that could have existed.
- Added some dumby projects into the deployTestRound script for ease of testing. Just uncomment those if you are using OptimisticUserRegistry.


corwintines:
Forgot to pick right project. Apologies.

PR #380, opened 2021-08-18
Feature sponsorship fee

auryn-macmillan:
This PR adds a `fee` and `feeRecipient` to the BrightIDUserRegistry, allowing contexts to set a fee that must be paid for self sponsorship. This mitigates the potential griefing attack in scenarios where the cost of a `sponsor()` call is lower than the cost of a sponsorship. The `fee` should be set to some value that is equal to or higher than the cost of a sponshorship. The `feeRecipient` could be set to the matching pool for a round, a funding source to the matching pool, or to some contract that buys sponsorships and allocates them to the same BrightID context.

### DIFF
--- a/contracts/contracts/userRegistry/BrightIdUserRegistry.sol
+++ b/contracts/contracts/userRegistry/BrightIdUserRegistry.sol
@@ -10,9 +10,13 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
     string private constant ERROR_NOT_AUTHORIZED = 'NOT AUTHORIZED';
     string private constant ERROR_INVALID_VERIFIER = 'INVALID VERIFIER';
     string private constant ERROR_INVALID_CONTEXT = 'INVALID CONTEXT';
+    string private constant ERROR_FEE_TO_LOW = 'FEE TO LOW';
 
     bytes32 public context;
     address public verifier;
+
+    uint public fee;
+    address public feeRecipient;
 
     struct Verification {
         uint256 time;
@@ -22,6 +26,8 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
 
     event SetBrightIdSettings(bytes32 context, address verifier);
     event Sponsor(address indexed addr);
+    event SetFee(uint fee);
+    event SetFeeRecipient(address feeRecipient);
 
     /**
      * @param _context BrightID context used for verifying users
@@ -38,8 +44,11 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
     /**
      * @notice Sponsor a BrightID user by context id
      * @param addr BrightID context id
+     * @notice Requires transaction value to be equal to fee.
      */
-    function sponsor(address addr) public {
+    function sponsor(address addr) public payable {
+        require(msg.value == fee, ERROR_FEE_TO_LOW);
+        payable(feeRecipient).transfer(msg.value);
         emit Sponsor(addr);
     }
 
@@ -55,6 +64,24 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
         context = _context;
         verifier = _verifier;
         emit SetBrightIdSettings(_context, _verifier);
+    }
+
+    /**
+     * @notice Set the sponsorship fee
+     * @param _fee Value to set the fee to
+    */
+    function setFee(uint _fee) external onlyOwner {
+        fee = _fee;
+        emit SetFee(fee);
+    }
+
+    /**
+     * @notice Set the address to receive fees
+     * @param _feeRecipient Address to receive fees
+    */
+    function setFee(address _feeRecipient) external onlyOwner {
+        feeRecipient = _feeRecipient;
+        emit SetFeeRecipient(feeRecipient);
     }
 
     /**
--- a/contracts/contracts/userRegistry/BrightIdUserRegistry.sol
+++ b/contracts/contracts/userRegistry/BrightIdUserRegistry.sol
@@ -81,7 +81,7 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
      * @notice Set the address to receive fees
      * @param _feeRecipient Address to receive fees
     */
-    function setFee(address _feeRecipient) external onlyOwner {
+    function setFeeRecipient(address _feeRecipient) external onlyOwner {
         feeRecipient = _feeRecipient;
         emit SetFeeRecipient(feeRecipient);
     }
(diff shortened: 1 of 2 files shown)


contracts/contracts/userRegistry/BrightIdUserRegistry.sol
```diff
@@ -57,6 +68,24 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
...
+    function setFee(uint _fee) external onlyOwner {
+        fee = _fee;
+        emit SetFee(fee);
+    }
+
+    /**
+     * @notice Set the address to receive fees
+     * @param _feeRecipient Address to receive fees
+    */
+    function setFee(address _feeRecipient) external onlyOwner {
```
pettinarip: `setFeeRecipient`?

vue-app/src/api/bright-id.ts
```diff
@@ -1,37 +1,44 @@
...
 ): Promise<TransactionResponse> {
-  const registry = new Contract(registryAddress, BrightIdUserRegistry, signer)
-  const userAddress = await signer.getAddress()
-  const transaction = await registry.sponsor(userAddress)
-  return transaction
+  const registry = new Contract(registryAddress, BrightIdUserRegistry, signer);
+  const userAddress = await signer.getAddress();
+  let overrides = {
+    value: await registry.fee
+  };
```
pettinarip, line 33: This `overrides` obj is not being used.

contracts/contracts/userRegistry/BrightIdUserRegistry.sol
auryn-macmillan: ah, yes!

vue-app/src/api/bright-id.ts
auryn-macmillan: lol, that was silly.
Thanks for reviewing.

contracts/contracts/userRegistry/BrightIdUserRegistry.sol
```diff
@@ -10,10 +10,14 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
     string private constant ERROR_NOT_AUTHORIZED = 'NOT AUTHORIZED';
     string private constant ERROR_INVALID_VERIFIER = 'INVALID VERIFIER';
     string private constant ERROR_INVALID_CONTEXT = 'INVALID CONTEXT';
+    string private constant ERROR_FEE_TO_LOW = 'FEE TO LOW';
```
samajammin:
```suggestion
    string private constant ERROR_FEE_TOO_LOW = 'FEE TOO LOW';
```

contracts/contracts/userRegistry/BrightIdUserRegistry.sol
```diff
@@ -38,8 +44,13 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
     /**
      * @notice Sponsor a BrightID user by context id
      * @param addr BrightID context id
+     * @notice Requires transaction value to be equal to fee.
      */
-    function sponsor(address addr) public {
+    function sponsor(address addr) public payable {
+        require(msg.value == fee, ERROR_FEE_TO_LOW);
```
samajammin:
```suggestion
        require(msg.value == fee, ERROR_FEE_TOO_LOW);
```

vue-app/src/api/bright-id.ts
```diff
@@ -1,37 +1,44 @@
...
+  signer: Signer
 ): Promise<TransactionResponse> {
-  const registry = new Contract(registryAddress, BrightIdUserRegistry, signer)
-  const userAddress = await signer.getAddress()
-  const transaction = await registry.sponsor(userAddress)
-  return transaction
+  const registry = new Contract(registryAddress, BrightIdUserRegistry, signer);
+  const userAddress = await signer.getAddress();
+  let overrides = {
+    value: await registry.fee
```
samajammin, line 32: Should this be `registry.fee()`? I believe storage variables are fetched with a getter.

contracts/contracts/userRegistry/BrightIdUserRegistry.sol
```diff
@@ -57,6 +68,24 @@ contract BrightIdUserRegistry is Ownable, IUserRegistry {
         emit SetBrightIdSettings(_context, _verifier);
     }
 
+    /**
+     * @notice Set the sponsorship fee
+     * @param _fee Value to set the fee to
+    */
+    function setFee(uint _fee) external onlyOwner {
```
samajammin, line 77: If we'd like, we could also set a default `fee` in the constructor, similar to how we set the `baseDeposit` in the OptimisticRecipientRegistry: https://github.com/clrfund/monorepo/blob/develop/contracts/contracts/recipientRegistry/OptimisticRecipientRegistry.sol#L64

samajammin (commented):
A couple minor comments. Overall looks good to me!

I think ideal scenario would be to have a fee paid in the ERC20 `nativeToken` but understand that increases the complexity quite a bit. This follows the same approach as the `baseDeposit` in the OptimisticRecipientRegistry 👍

samajammin:
Another general note - I'm noticing we don't have any tests for the BrightIDUserRegistry, only the SimpleUserRegistry. We should probably write tests at some point 😅.

auryn-macmillan:
> I think ideal scenario would be to have a fee paid in the ERC20 `nativeToken` but understand that increases the complexity quite a bit. This follows the same approach as the `baseDeposit` in the OptimisticRecipientRegistry

Yeah, I decided to go this route because it means that we don't need to modify the user flow at all. If we do it with ERC20's then it means extra steps for the user. The downside with this pattern, though, is that you can't send the funds directly to the matching pool.

> Another general note - I'm noticing we don't have any tests for the BrightIDUserRegistry, only the SimpleUserRegistry. We should probably write tests at some point sweat_smile.

Ha ha, yeah I noticed this as well. I was in a rush last week, so didn't have time to write out tests. But I'll go ahead and merge this now and write out some tests for it later this week.

contracts/contracts/userRegistry/BrightIdUserRegistry.sol
auryn-macmillan: Ha ha! Good catch.

daodesigner:
Adding making the changes suggested and writing some tests to the next sprint unless I hear otherwise. @auryn-macmillan are we dropping this feature ?

auryn-macmillan:
> Adding making the changes suggested and writing some tests to the next sprint unless I hear otherwise. @auryn-macmillan are we dropping this feature ?

I still think this feature makes sense, but it's not a critical one, so we can prioritize accordingly.

yuetloo:
Closing this PR as it's old and probably obsolete. If we need it again, we will create a new PR.

Issue #359, opened 2021-05-20
Use mastercopy-proxy pattern for all factory contracts


auryn-macmillan:
Currently, the `FundingRoundFactory` deploys a new instance of `FundingRound` for each new round (which also deploys an instance of MACI). This transaction costs ~10,000,000 gas.

We can dramatically reduce this gas cost by implementing a pattern similar to the Gnosis Safe's mastercopy-proxy pattern, where each version of the contracts has a reference deployment and any [factory contracts deploy proxies](https://github.com/gnosis/safe-contracts/tree/main/contracts/proxies) that make delegate calls to those reference deployments.

This will enable:
1. Deployments that are orders of magnitude cheaper
2. Persistent addresses for FundingRoundFactory contracts (they can be upgraded between rounds, no need to redeploy)
3. Anyone to [replay the deployment](https://github.com/gnosis/safe-contract-deployment-replay) of the reference implementations of the contracts to any network.

auryn-macmillan:
@daodesigner, @viraj124 is keen to take on this issue. Just wanted to touch base with you on it since it will likely impact (and may have already been addressed by) your work on the deployer.

daodesigner:
That would be great, should be a drop in replacement on the deployer side, so this would be a great first issue to tackle. Let's coordinate to use the same contracts for subgraph registration/agree on an ABI.

viraj124:
Thanks @daodesigner Will coordinate with you when I start

daodesigner:
For reference here are the current deploy costs for each of the contracts after compiler optimization. 
*note: the deployNewRound contract call deploys `FundingRound`/`MACI` contracts.

```
Contract deployment: MACIFactory
Gas used:            5,178,462 

Contract deployment: FundingRoundFactory
Gas used:            4,555,109 

Contract deployment: OptimisticRecipientRegistry
Gas used:            1,591,758 

Contract deployment: BrightIdUserRegistry
Gas used:            702,831 

Contract call:       FundingRoundFactory#deployNewRound
Gas used:            7,749,574 
```

viraj124:
@daodesigner I will probably start on this on weekend will keep you posted
Thanks

viraj124:
@daodesigner currently deployment takes place via a function as you said so you expect this behavior to remain the same so it does not affect the current deployment flow ?
[...]

PR #385, opened 2021-09-21
OZ Proxy for MACI Factory with test coverage

viraj124:
https://github.com/clrfund/monorepo/issues/359

This PR adds the proxy for the MACI Factory with unit test coverage in addition to the Funding Round Factory Proxy added by @daodesigner

### DIFF
--- a/contracts/contracts/ClrFund.sol
+++ b/contracts/contracts/ClrFund.sol
@@ -45,17 +45,18 @@ contract ClrFund is OwnableUpgradeable, MACISharedObjs {
   event CoordinatorChanged(address _coordinator);
 
    // constructor to prevent anyone anonymous getting the ownership
-  constructor(MACIFactory _maciFactory) public {
-    require(address(_maciFactory) != address(0), "ClrFund: invalid address");
-    init(_maciFactory);
+  constructor(MACIFactory _maciFactory, address _owner) public {
+    init(_maciFactory, _owner);
   }
 
   function init(
-    MACIFactory _maciFactory
+    MACIFactory _maciFactory,  address _owner
   ) 
     public
   {
+    // initial ownership set is msg.sender so the ownership needs to be transferred in the init method
     __Ownable_init();
+    transferOwnership(_owner);
     maciFactory = _maciFactory;
   }
 
--- a/contracts/contracts/ClrFund.sol
+++ b/contracts/contracts/ClrFund.sol
@@ -44,10 +44,16 @@ contract ClrFund is OwnableUpgradeable, MACISharedObjs {
   event TokenChanged(address _token);
   event CoordinatorChanged(address _coordinator);
 
+   // constructor to prevent anyone anonymous getting the ownership
+  constructor(MACIFactory _maciFactory) public {
+    require(address(_maciFactory) != address(0), "ClrFund: invalid address");
+    init(_maciFactory);
+  }
+
   function init(
     MACIFactory _maciFactory
   ) 
-    external
+    public
   {
     __Ownable_init();
     maciFactory = _maciFactory;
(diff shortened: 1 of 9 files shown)


auryn-macmillan (approved):
LGTM -- @daodesigner any issues with merging this?

daodesigner:
Yeah apologies for the delay here, only recently had time to do a serious review. @viraj124 please add constructors to initialize the singletons (also do this for the initial proxy contract I wrote please) and write tests to verify this.

viraj124:
> Yeah apologies for the delay here, only recently had time to do a serious review. @viraj124 please add constructors to initialize the singletons (also do this for the initial proxy contract I wrote please) and write tests to verify this.

@daodesigner  I think they are already there https://github.com/viraj124/monorepo/blob/subgraph/contracts/contracts/ClrFundMACIFactoryDeployer.sol#L34 probably the test cases are missing I will check that

daodesigner:
hey just closing the loop here: I mean a constructor on the singleton(s) (template that gets passed in initially to the proxy factory)

contracts/contracts/ClrFundMACIFactory.sol
```diff
@@ -0,0 +1,148 @@
...
+  SnarkVerifier public batchUstVerifier;
+  SnarkVerifier public qvtVerifier;
+  uint256 public signUpDuration;
+  uint256 public votingDuration;
+
+  // Events
+  event MaciParametersChanged();
+  event MaciDeployed(address _maci);
+
+  function init(
```
auryn-macmillan, line 60: For the avoidance of doubt, @viraj124, @daodesigner is suggesting (as I suggested in our DMs a while back) that this contract should have a constructor that calls your `init()` function.
This way you can:
1. choose how you want to deploy the contract, directly or via proxy.
2. deploy and immediately initialize the singleton.

viraj124: yeah got that @auryn-macmillan just updating the test cases

viraj124:
@daodesigner @auryn-macmillan pushed the changes

contracts/contracts/ClrFund.sol
```diff
@@ -44,10 +44,16 @@ contract ClrFund is OwnableUpgradeable, MACISharedObjs {
   event TokenChanged(address _token);
   event CoordinatorChanged(address _coordinator);
 
+   // constructor to prevent anyone anonymous getting the ownership
+  constructor(MACIFactory _maciFactory) public {
+    require(address(_maciFactory) != address(0), "ClrFund: invalid address");
```
auryn-macmillan: I think these checks are unnecessary.
It's unlikely that anyone passes `address(0)` in as an argument, and if they do they only bork their own deployment, it doesn't do any damage to anyone else.

contracts/contracts/ClrFund.sol
```diff
@@ -44,10 +44,16 @@ contract ClrFund is OwnableUpgradeable, MACISharedObjs {
   event TokenChanged(address _token);
   event CoordinatorChanged(address _coordinator);
 
+   // constructor to prevent anyone anonymous getting the ownership
+  constructor(MACIFactory _maciFactory) public {
```
auryn-macmillan: Given that the contracts are now ownable, the constructor should allow you to pass in `owner` as a parameter, rather than assuming msg.sender is the owner.

auryn-macmillan (changes requested):

contracts/contracts/ClrFund.sol
viraj124: Just added to be very safe will remove 
Thanks

contracts/contracts/ClrFund.sol
viraj124: Sure, 
Thanks

auryn-macmillan (approved):
LGTM. @daodesigner, this all check out?

PR #413, opened 2021-12-14
Blank banner during join phase

### LINKED ISSUE #463: Blank banner during join phase
If there's no current round, the RoundStatusBanner is blank:
Image 2021-12-08 at 3 29 58 PM

For this case we should just hide the banner / not display it at all.

corwintines:
Fixes: https://github.com/ethereum/clrfund/issues/463

**What Changed**
- Add conditional to only render banner if there is a round


corwintines:
Sorry, meant to PR to our fork. Bad habit.

PR #462, opened 2022-03-01
Fix/contract test cases

yuetloo:
This PR fixes the test failures due to `removeRecipient()` expecting a baseDeposit.
It also updated the test case for `allow only owner to execute remove recipient request during challenge period`


daodesigner (approved):
lgtm

PR #472, opened 2022-03-07
Make `rounds` array public

auryn-macmillan:
Make the `rounds` array public so that it can be easily queried by an RPC node.


samajammin:
This is certainly an option! Thought it may not be necessary. We have an API method `getRounds` which queries this data via the subgraph: https://github.com/clrfund/monorepo/blob/develop/vue-app%2Fsrc%2Fapi%2Frounds.ts#L10

So I believe this is already achievable. @pettinarip @daodesigner correct me if I'm wrong here.

I suppose it depends on the use case, i.e. if there's instances we need to be able to query the contract directly.

PR #474, opened 2022-03-08
Add VUE_APP_FIRST_ROUND to ignore old rounds

samajammin:
- Add `VUE_APP_FIRST_ROUND` to make it possible to ignore previous rounds, triggering the app into the "join phase"
- Remove the /rounds route, as it's not needed in this round

### DIFF
--- a/vue-app/src/api/round.ts
+++ b/vue-app/src/api/round.ts
@@ -58,7 +58,7 @@ export async function getCurrentRound(): Promise<string | null> {
     isSameAddress(round.address, fundingRoundAddress)
   )
 
-  if (roundIndex >= Number(process.env.VUE_APP_FIRST_ROUND)) {
+  if (roundIndex >= Number(process.env.VUE_APP_FIRST_ROUND || 0)) {
     return fundingRoundAddress
   }
   return null
--- a/vue-app/src/api/round.ts
+++ b/vue-app/src/api/round.ts
@@ -5,6 +5,9 @@ import { PubKey } from 'maci-domainobjs'
 import { FundingRound, MACI, ERC20 } from './abi'
 import { provider, factory } from './core'
 import { getTotalContributed } from './contributions'
+import { getRounds } from './rounds'
+
+import { isSameAddress } from '@/utils/accounts'
 
 export interface RoundInfo {
   fundingRoundAddress: string
@@ -50,7 +53,15 @@ export async function getCurrentRound(): Promise<string | null> {
   if (fundingRoundAddress === '0x0000000000000000000000000000000000000000') {
     return null
   }
-  return fundingRoundAddress
+  const rounds = await getRounds()
+  const roundIndex = rounds.findIndex((round) =>
+    isSameAddress(round.address, fundingRoundAddress)
+  )
+
+  if (roundIndex >= Number(process.env.VUE_APP_FIRST_ROUND)) {
+    return fundingRoundAddress
+  }
+  return null
 }
 
 //TODO: update to take factory address as a parameter, default to env. variable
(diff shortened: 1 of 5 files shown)


samajammin:
@pettinarip please take a look - I tried to recreate the implementation you showed me.

vue-app/src/api/round.ts
```diff
@@ -50,7 +53,15 @@ export async function getCurrentRound(): Promise<string | null> {
   if (fundingRoundAddress === '0x0000000000000000000000000000000000000000') {
     return null
   }
-  return fundingRoundAddress
+  const rounds = await getRounds()
```
pettinarip, line 56: `getRounds` is not returning the data in the correct order. We need to sort it first. Will push those changes

samajammin:
I tested this against the current Arbitrum One deploy & it appears to work as expected (by setting `VUE_APP_FIRST_ROUND` ENV variable to `2`). @auryn-macmillan I'm seeing you've deployed 2 round with the `0x12F85B7633C0c3E38A892C78d081Ba4818214Eff` factory, is that correct? If not, something funky is going on with the `getRounds` api method @pettinarip as it's returning 2 results:
1. https://arbiscan.io/address/0x3aca64bff8b6501c740a7fbbe8c794f7cb2074dc (which is said not to exist)
2. https://arbiscan.io/address/0xc6aefed2ec944547dbe1812474fd8286f65b64b3 (looks like the actual 1st round)

vue-app/src/graphql/API.ts
```diff
@@ -2505,7 +2505,7 @@ export const GetRecipientsDocument = gql`
     `;
 export const GetRoundsDocument = gql`
     query GetRounds {
-  fundingRounds {
+  fundingRounds(orderBy: startTime, orderDirection: asc) {
```
pettinarip, line 2508: I'm using the `startTime` since we don't have an `index` in the entity to replicate the exact order of the contract's rounds array.

pettinarip:
> @auryn-macmillan I'm seeing you've deployed 2 round with the `0x12F85B7633C0c3E38A892C78d081Ba4818214Eff` factory, is that correct? If not, something funky is going on with the `getRounds` api method @pettinarip as it's returning 2 results:

Maybe we had a failed deployment? not sure how that could happen

pettinarip:
I see two `deployNewRound` txs
https://arbiscan.io/tx/0x4d8c8f0fe31df2f1bf6f5a8112ce726566a9b0c557690b4073babbbe449a0ac2
https://arbiscan.io/tx/0xd43dc9633c15149ff262b639bb2fe6a9d096a03bb036f695207dfda07a8fa3a3
They seem to be ok.
We can wait for Auryn confirmation, but it seems that he deployed 2 rounds for testing I guess.

pettinarip (approved):

vue-app/src/api/round.ts
```diff
@@ -50,7 +53,15 @@ export async function getCurrentRound(): Promise<string | null> {
...
+  const roundIndex = rounds.findIndex((round) =>
+    isSameAddress(round.address, fundingRoundAddress)
+  )
+
+  if (roundIndex >= Number(process.env.VUE_APP_FIRST_ROUND)) {
```
pettinarip: We can default it to `0`. 

```suggestion
  if (roundIndex >= Number(process.env.VUE_APP_FIRST_ROUND || 0)) {
```

samajammin: The empty string automatically defaults to 0 but I'm good with making that more explicit 👍

auryn-macmillan:
Just to confirm, yes I deployed two test rounds.

PR #475, opened 2022-03-09
undefined chain.bridge error for non layer2

yuetloo:
When running the app on rinkeby, noticed the error in console log about undefined props for links.

The fix will remove the "get fund" button for non-layer2.

For layer2, it will redirect to the about layer2 page.. Not sure if it should go directly to the layer2 bridge or the about page.

Here's how it looks on the landing page.
Screen Shot 2022-03-08 at 10 04 33 PM

### DIFF
--- a/vue-app/src/views/Landing.vue
+++ b/vue-app/src/views/Landing.vue
@@ -93,10 +93,7 @@
               <b>{{ chain.label }} for fast and cheap transaction fees</b>
             </p>
           </div>
-          <links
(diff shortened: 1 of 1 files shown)


auryn-macmillan:
@yuetloo, I thought about this more and realized that it's probably not possible to generalize the content at the `/#/about/layer-2` route in a way that would make it actually useful on any network. So I'm in favour of just removing this route for networks that are not arbitrum. Going to go ahead and merge this PR rather than creating a new issue to generalize.

PR #486, opened 2022-04-24
User registry factory when no round exists

pettinarip:
Read the user factory address from the factory when there is no current round.


samajammin (approved):
👍

PR #487, opened 2022-05-05
Add optional ENV variable to customize matching pool

samajammin:
Request from @auryn-macmillan 

> I'd like to change it so that when people hit the button to contribute funds to the matching pool, it goes to the matching pool safe rather than the funding round factory.
That way, if for some reason we have a hiccup in the round, we don't have to do any wrangling to recover those funds.

This provides an optional ENV variable that would override the address where contributions to the matching pool are transferred to.

### DIFF
--- a/vue-app/src/api/round.ts
+++ b/vue-app/src/api/round.ts
@@ -154,14 +154,6 @@ export async function getRoundInfo(
     matchingPool = await factory.getMatchingFunds(nativeTokenAddress)
   }
 
-  // Hack to overwrite matching pool amount using designated address
-  if (nativeTokenAddress && process.env.VUE_APP_MATCHING_POOL_ADDRESS) {
-    matchingPool = await getTokenBalance(
-      nativeTokenAddress,
-      process.env.VUE_APP_MATCHING_POOL_ADDRESS
-    )
-  }
-
   const totalFunds = matchingPool.add(contributions)
 
   return {
--- a/vue-app/src/api/round.ts
+++ b/vue-app/src/api/round.ts
@@ -8,7 +8,6 @@ import { getTotalContributed } from './contributions'
 import { getRounds } from './rounds'
 
 import { isSameAddress } from '@/utils/accounts'
-import { getTokenBalance } from '@/api/user'
 
 export interface RoundInfo {
   fundingRoundAddress: string
--- a/vue-app/src/api/round.ts
+++ b/vue-app/src/api/round.ts
@@ -8,6 +8,7 @@ import { getTotalContributed } from './contributions'
 import { getRounds } from './rounds'
 
 import { isSameAddress } from '@/utils/accounts'
+import { getTokenBalance } from '@/api/user'
 
 export interface RoundInfo {
   fundingRoundAddress: string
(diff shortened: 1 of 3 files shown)


samajammin:
@auryn-macmillan I believe this accomplishes our needs but my concern here is that the UI will no longer update the amount in the matching pool correct, since it queries the factory/current round addresses for that info. I think that would require an update to the subgraph 😕 

edit: to overcome this, perhaps we could update how we query that info in the UI to just query the token balance of that address vs. relying on the subgraph like we currently do. That would miss any contributions to the current funding round contract or the funding round factory contract but perhaps that doesn't matter since no one would be sending funds to those contracts (at least not through the UI).

auryn-macmillan:
> my concern here is that the UI will no longer update the amount in the matching pool correct

Does the subgraph not currently account for external funding sources?

samajammin:
> Does the subgraph not currently account for external funding sources?

That's a great question. Not sure but can look into it or maybe @daodesigner knows?

I just pushed a commit which I believe will solve the UI issue. It appears we weren't actually relying on the subgraph for this data yet anyway.

daodesigner:
> > Does the subgraph not currently account for external funding sources?
> 
> 
> 
> That's a great question. Not sure but can look into it or maybe @daodesigner knows?
> 
> 
> 
> I just pushed a commit which I believe will solve the UI issue. It appears we weren't actually relying on the subgraph for this data yet anyway.

The subgraph does account for the matching pool at the moment by calling the [getMarchingFunds method](https://github.com/clrfund/monorepo/blob/48857be277be187f1b9b63643c7ebd24c1b862e3/contracts/contracts/ClrFund.sol#L185) on the contract. However at the moment this is only updated when a new round starts and ends in [this subgraph handler](https://github.com/clrfund/monorepo/blob/48857be277be187f1b9b63643c7ebd24c1b862e3/subgraph/src/FundingRoundFactoryMapping.ts#L97). Lmk if you guys need it to be refreshed more often.

samajammin:
Ah yes @daodesigner thanks for the refresher. Been a few months, amazed by how quickly I've forgotten the codebase 🤦 Admittedly pulled this together in a bit of a rush.

So in that case, if the safe has been added as a funding source, I believe we should be ok given that the round info queries the `getMatchingFunds` method from the factory contract as @daodesigner pointed out:
https://github.com/clrfund/monorepo/blob/cohort/EthStaker/vue-app/src/api/round.ts#L153
If this is correct, I should remove the extra logic I added to overwrite the `matchingPool` amount in /api/round.ts

From what I see, none of the UI relies on the subgraph for this - is that correct?

@auryn-macmillan could you please confirm if the safe has been added as a funding source?

auryn-macmillan:
> @auryn-macmillan could you please confirm if the safe has been added as a funding source?

@samajammin, yeah I can confirm it's already been added as a funding source.

> However at the moment this is only updated when a new round starts and ends

I think this will probably work in most cases. So I don't think there is any great urgency to update it. The caveat being when additional funds are added to a funding source safe during the round or the allowances are updated during a round.

For this round, we set unlimited approval from the matching pool safe to the funding round factory, so any Dai deposited to the safe will be transferred at the end of the round. That being the case, it seems like the total balance of the matching pool (across all funding sources and the funding round factory) can probably be best fetched via RPC calls, and perhaps the subgraph should just track the list of funding sources. The app would query the balanceOf and allowance of each of those addresses, and then sum those to display the total matching pool.

vue-app/src/api/round.ts
```diff
@@ -153,6 +154,14 @@ export async function getRoundInfo(
     matchingPool = await factory.getMatchingFunds(nativeTokenAddress)
   }
 
+  // Hack to overwrite matching pool amount using designated address
```
auryn-macmillan: Ideally, we should query both balanceOf and allowance on each fundingSource, and sum the lower of the two for each funding source. Not necessary in this case though.

samajammin:
> If this is correct, I should remove the extra logic I added to overwrite the matchingPool amount in /api/round.ts

Ok @auryn-macmillan I've removed that extra logic, so now we continue to rely on the `getMatchingFunds` method from the factory contract.

I believe this is now good to go 👍👍

Issue #448, opened 2022-01-26
Generalize how round info is fetched on app load


pettinarip:
**_ISSUE MIGRATED FROM_** https://github.com/ethereum/clrfund/issues/447 _(check it for more context) originally created by_ @samajammin

***

Currently the data fetching of the app is built specifically for the Eth2 round use case - it essentially assumes we only care about the current round.

In App.vue we fetch the current round as soon as this root component is mounted: https://github.com/ethereum/clrfund/blob/develop/vue-app/src/App.vue#L106
We do this so we can load round information as soon as possible, e.g. in the [RoundStatusBanner](https://github.com/ethereum/clrfund/blob/develop/vue-app/src/components/RoundStatusBanner.vue) which we display on the landing page:
![Image 2021-10-05 at 2 03 06 PM](https://user-images.githubusercontent.com/8097623/136102441-7638ecbe-4e53-4fd7-81e9-9221e8f9ce62.jpg)

This all works well but we run into potential issues with the ProjectList component if the user wants to view historical rounds, e.g. via the rounds route: https://clr.fund/#/rounds. Selecting a round adds the round address to the route param (e.g. https://clr.fund/#/round/0xf8acacfA954742Dc4eaf8Bd8498F4DFdc01B3875) which ProjectList.vue uses to load the round info: https://github.com/ethereum/clrfund/blob/develop/vue-app/src/views/ProjectList.vue#L151
This could result in clashing data - ProjectList.vue would set the round using the route param but the polling in App.vue may overwrite this if it fetches `getCurrentRound`.

We should refactor this to avoid any potential clashes, e.g. by ensuring that App.vue only sets the round address to the current round if there is no address in the route params.

PR #490, opened 2022-05-10
Merge features exclude content

daodesigner:


contracts/contracts/recipientRegistry/OptimisticRecipientRegistry.sol
```diff
@@ -177,16 +177,11 @@ contract OptimisticRecipientRegistry is Ownable, BaseRecipientRegistry {
     */
   function executeRequest(bytes32 _recipientId)
     external
+    onlyOwner
     returns (bool)
```
yuetloo, line 181: @auryn-macmillan , do we want this change in the `develop` branch?  This looks like ethStaker specific change.

netlify.toml
```diff
@@ -0,0 +1,12 @@
+[build]
+  command = "yarn build"
+  functions = "vue-app/dist/lambda"
+  publish = "vue-app/dist"
+
+[context.staking-round.environment]
+  VUE_APP_ETHEREUM_API_URL = "https://arb-rinkeby.g.alchemy.com/v2/wPwZJEp0lIu0XwccG_t0MAqF-phtI3og"
+  VUE_APP_ETHEREUM_API_CHAINID = "421611"
+  VUE_APP_CLRFUND_FACTORY_ADDRESS = "0x9cF8c20dAb0aAFC6584D5498C0A97F0F51D6F7E4"
+  VUE_APP_SUBGRAPH_URL = "https://api.thegraph.com/subgraphs/name/pettinarip/clrfundrinkarby2"
```
yuetloo, line 10: @auryn-macmillan should we exclude this config file from the `develop` branch?

contracts/contracts/recipientRegistry/OptimisticRecipientRegistry.sol
yuetloo: @auryn-macmillan , I removed this change anyways as it's failing unit tests. If you want this change, I can put it back and fix the test cases instead.

contracts/contracts/snarkVerifiers/QuadVoteTallyVerifier32Batch16.sol
```diff
@@ -0,0 +1,240 @@
+// SPDX-License-Identifier: MIT
+
+// Copyright 2017 Christian Reitwiessner
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
```
yuetloo, line 6: @auryn-macmillan , I'm keeping these verifier contracts, even though they are not used, just in case for future reference.

contracts/contracts/recipientRegistry/OptimisticRecipientRegistry.sol
auryn-macmillan: Yeah, this was an ETHStaker specific change.

We could perhaps add this back in as a separate registry type in future.

netlify.toml
auryn-macmillan: Yes, I think so. Do you agree?

vue-app/src/router/index.ts
```diff
@@ -59,14 +58,6 @@ const routes = [
       title: 'Round Information',
     },
   },
-  {
-    path: '/rounds',
-    name: 'rounds',
-    component: RoundList,
-    meta: {
-      title: 'Rounds',
-    },
```
yuetloo, line 68: @auryn-macmillan I just realized the reason the rounds endpoint was removed is related to #448 and I'm still assessing the impact of the issue.  So far i've found out that regardless of the round selected, the app always shows `current round` information because of the reset logic in the app.vue.  Maybe we should even remove the /round/:id route as it has the same issue

contracts/contracts/recipientRegistry/OptimisticRecipientRegistry.sol
yuetloo: I added EthStaker's contract as `PermissionedRecipientRegistry` for now. But, the UI will still treat it as "optimistic".

vue-app/src/router/index.ts
auryn-macmillan: Removing the `/round/:id` route would mean that users cannot view previous rounds, correct?

In an ideal world, I would prefer for the app to be able to display previous rounds. However, in practice this has turned out to be quite difficult, since previous rounds have often used different deployments of the contracts, earlier versions of the contracts, different configurations of contracts (recipient registries, etc), and/or been deployed to different networks. All of which is a huge pain to track in the context of one app.

Perhaps a cleaner solution is to treat each round as independent app and simply move previous rounds to a subdomain, like `round1.clr.fund`, when we begin a new round?

But, if we're confident that most people will not change networks and versions as often as we have (probably a reasonable assumptions), then it would be ideal for the app to support multiple rounds.

yuetloo: Correct, removing the `/round/:id` route will prevent users from viewing previous rounds.

Yeah, not sure how well the app supports cross-chain round viewing.  It does support loading external previous rounds by setting the ipfs hash in VUE_APP_EXTRA_ROUNDS. 

See the below screenshot, for external rounds, they have a little arrow beside the round. Clicking the round will open the ipfs url for the app hosting the round. So, this is similar to your solution, host previous rounds in a subdomain.

I would also like to get your opinion on whether the round information panel (the left panel)  should always display the current round information or the selected round information depending on which page you're on?

/rounds:
Screen Shot 2022-07-07 at 10 13 27 AM

/projects
Screen Shot 2022-07-07 at 10 14 59 AM

/round/:id
Screen Shot 2022-07-07 at 10 14 45 AM

vue-app/src/api/contributions.ts
```diff
@@ -5,15 +5,16 @@ import { TransactionResponse } from '@ethersproject/abstract-provider'
...
+import { FundingRound, ERC20 } from './abi'
+import { factory, provider } from './core'
 import { Project } from './projects'
 import sdk from '@/graphql/sdk'
 
 export const DEFAULT_CONTRIBUTION_AMOUNT = 5
 export const MAX_CONTRIBUTION_AMOUNT = 10000 // See FundingRound.sol
 
 // The batch of maximum size will burn 9100000 gas at 700000 gas per message
-export const MAX_CART_SIZE = 13
```
yuetloo, line 16: Should we make this parameter configurable and throw if it's not configured?  This is to make sure we tested the max size before deploying to production.

yuetloo: @auryn-macmillan

vue-app/src/router/index.ts
auryn-macmillan: > See the below screenshot, for external rounds, they have a little arrow beside the round. Clicking the round will open the ipfs url for the app hosting the round. So, this is similar to your solution, host previous rounds in a subdomain.

Yeah, that's a good point. I think it's reasonable to just leave it as is for now.

> I would also like to get your opinion on whether the round information panel (the left panel) should always display the current round information or the selected round information depending on which page you're on?

My preference is for it to display the information on the selected round.

PR #491, opened 2022-05-10
Fix condition to check if the round has started

pettinarip:
We are displaying (in https://qf.ethstaker.cc/#/verify) the following message incorrectly since the round has already started.
bug

This was because we were assuming that the round was not started if we were in the join phase. But we can be in the join phase while the round is open.

To check if the round has started we just check if there is a current round.


samajammin (approved):
Nice catch!

PR #496, opened 2022-05-16
only render brightid link if available

yuetloo:
This PR fixes the 404 page not found error when the brightid link is clicked in the mobile mode.

The link is rendered before the `appLink` is set, so, it's rendered as `router-link` instead of an external link causing the page not found error.


daodesigner:
Lgtm

daodesigner (approved):

PR #497, opened 2022-05-20
Fetch round contributions balance from the contract

pettinarip:
We were fetching each individual contribution from thegraph endpoint (~1800 contributions so far).

Seems that thegraph has understandable limits on the amount of data they return (100 items).

This PR fetches the balance from the contract directly instead of thegraph endpoint.


auryn-macmillan (approved):
lgtm

PR #527, opened 2022-06-30
Better bright id verification UX

### LINKED ISSUE #438: Make verification check more clear
[ THIS IS A MIGRATED ISSUE - ORIGINAL: https://github.com/ethereum/clrfund/issues/330 ]

Just walked through the verification flow, which involved clicking "Check if you are verified" to view my BrightID verification status:

https://user-images.githubusercontent.com/8097623/131523338-cb4ca481-50aa-46e6-b26a-11239101c3c4.mp4

I don't think this UX is currently clear that the check is working. It flashes so quickly that it seems like it might be a bug.

Perhaps we can add additional text or refresh the "Unverified" text e.g. with a CSS animation to show that we did successfully complete a check & the result is that they're still not yet verified.

### LINKED ISSUE #506: Make the `Set up BrightId` page show brightId's error
When a user failed to verify by `BrightId`, the following page will display, and clicking the `Check if you are verified` button will display the same page without a new error message.  Displaying brightid errors on the page will be very helpful for troubleshooting.

One way to reproduce this error is to verify different accounts on Brightid and log in using the older account.  In this case the brightId api will return the account as non-unique.

Screen Shot 2022-06-14 at 9 54 40 AM

yuetloo:
this fixes #506 and #438

### DIFF
--- a/vue-app/src/views/Verify.vue
+++ b/vue-app/src/views/Verify.vue
@@ -91,14 +91,34 @@
                   class="btn-primary"
                   :disabled="loadingManualVerify"
                 >
-                  Check if you are verified
+                  <loader v-if="loadingManualVerify" class="button-loader" />
+                  <span v-else>Check if you are verified</span>
                 </button>
-                <loader v-if="loadingManualVerify" />
+                <div v-if="showVerificationResult" class="error">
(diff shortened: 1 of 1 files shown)


auryn-macmillan (approved):
lgtm 👍

PR #529, opened 2022-07-07
Merge ethstaker features (excluding content)

yuetloo:
Create this as new PR to replace [ PR#490 ](https://github.com/clrfund/monorepo/pull/490) to avoid updating https://qf.ethstaker.cc/#/


auryn-macmillan:
@yuetloo, there are a handful of places in the app where `clr.fund` is mentioned in the copy. In most cases, where it is talking about the `operator` or title of the round, rather than the application itself, I think we should probably change the copy to be variables that are settable in the .env file, perhaps they default to clr.fund if not set.

Thoughts?

yuetloo:
> @yuetloo, there are a handful of places in the app where `clr.fund` is mentioned in the copy. In most cases, where it is talking about the `operator` or title of the round, rather than the application itself, I think we should probably change the copy to be variables that are settable in the .env file, perhaps they default to clr.fund if not set.
> 
> Thoughts?

Make sense, I'll update it

auryn-macmillan (approved):
lgtm

PR #532, opened 2022-07-14
fix breadcrumb alignment

yuetloo:
Fixes breadcrumbs alignment again because the fix in PR #520 broke the alignment on /projects.


auryn-macmillan (approved):
lgtm

PR #473, opened 2022-03-07
Generalize recipient registration flow

yuetloo:
This PR closes #447 

UI changes:

1) If the recipient registry type is not `optimistic`, accessing the recipient registry page will show the spinner (current behaviour).  This is changed to show the "Not Found" page
recipient-registry

2) Round criteria - removed the "Add project" button on the bottom of the page. Should there be a button or link to instruct users to contact operator to add their projects? 
Screen Shot 2022-03-10 at 9 36 33 AM

3) Join Landing page - removed the "Add project" button
Screen Shot 2022-03-10 at 9 36 26 AM

4) About recipient - removed the deposit requirement if registry is not optimistic, but left the link to the "join" landing page.  Should there be instruction to users to contact the operator to add their project?
Screen Shot 2022-03-10 at 9 35 26 AM

5) Project page - should the "Add project" button be removed from the page for "non-optimistic" registry? I left it because it currently redirects to the "join landing" page which has links to the "Round Criteria"
Screen Shot 2022-03-10 at 9 36 55 AM

6) Landing page - should the "Add project" button be removed here? I left it because it redirects to the "join landing" page
Screen Shot 2022-03-10 at 10 18 27 AM

PR #538, opened 2022-08-03
Allow recipient registry owner to add recipients for simple recipient registry type

### LINKED ISSUE #478: Subgraph mapping for recipient removal for simple/kleros recipient registry
This issue was introduced by PR https://github.com/clrfund/monorepo/pull/473

The issue is to revisit the definition of the recipient entity in subgraph.

Currently, the "removeRecipient" mapping rule marked recipient as 'rejected' in the subgraph when the recipient was deleted from the contract.  (note: this logic only applies to simple and Kleros registry, optimistic registry has different logic).

See RecipientMapping.ts, removeRecipient() function for the mark 'rejected' code.

Also, in the function getRegistryInfo() in file /api/recipient-registry.ts, the logic gets the recipientCount from the registry contract instead of from the subgraph because it cannot simply count the recipient records. The recipient entity in the subgraph does not store recipient entity. Instead it stores the operation( requests ) on recipients generated from the optimistic recipient registry.

Screen Shot 2022-03-10 at 11 09 11 AM

Will leave this issue to be tackled at a later time because the UI code that depends on the information is currently retrieving information that it needs from the contract directly as opposed to from the subgraph.  This is to keep the scope of the original PR small.

### LINKED ISSUE #510: Add support for multiple registry types
#473 details and addresses some of the issues in the app if a deployment is using something other than the `optimistic` recipient registry. Ideally, we should generalized the app so that it supports the `optimistic`, `simple`, and `kleros` registries. Would should consolidate this with the [effort to](https://github.com/clrfund/monorepo/tree/feature/metadata) use the [metadata registry](https://github.com/clrfund/metadata-registry).

We should:

- [ ] replace `_metadata` with `_metadataId` in each of the registries
- [ ] pull all recipient metadata from the metadata registry
- [ ] remove the `/#/join` route and any copy, links, and CTAs to it when the `simple` registry is in use.
- [ ] build native support for the `kleros` registry, with the assumption that all metadata should be stored in the metadata registry. (should resolve #480)

yuetloo:
This fixes the first 3 items in #510 

@auryn-macmillan , please review the changes and let me know if the registration flow is acceptable because for simple registry, once owner added the recipient, it automatically shows up, no need to go to the /recipients page to accept the recipient.  Also, the UI currently doesn't support deleting a recipient from the simple registry atm.

Also, should we revise the content for `/about/recipient` since it's applicable to the optimistic registry only?  I did attempt to remove the optimistic wording in this [PR](https://github.com/clrfund/monorepo/pull/473) which will be deprecated in favour of this PR.

Screen Shot 2022-08-03 at 12 38 55 PM

### DIFF
--- a/contracts/scripts/deployRecipientRegistry.ts
+++ b/contracts/scripts/deployRecipientRegistry.ts
@@ -2,10 +2,6 @@ import { ethers } from 'hardhat'
 import { UNIT } from '../utils/constants'
 import { RecipientRegistryFactory } from '../utils/recipient-registry-factory'
 
-function isSimpleRegistry(registryType: string): boolean {
-  return registryType === 'simple'
-}
-
 /*
  * Deploy a new recipient registry.
  * The following environment variables must be set to run the script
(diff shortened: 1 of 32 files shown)


auryn-macmillan (approved):
lgtm :tada:

PR #539, opened 2022-08-03
Add email form for recipient registration

### LINKED ISSUE #516: Add a new step to accept recipient email for the recipient registration process
Problem:
If the environment variable, VUE_APP_GOOGLE_SPREADSHEET_ID, is defined, the optimistic recipient registration process will require the user to input an email address, and the email address is not recorded in the recipient registry contract.

This flow is broken with the introduction of the universal recipient registry which uses recipient metadata from an external source. Because the email is not available in the recipient metadata, a new page/step is needed in the recipient registration process to take the user email address when VUE_APP_GOOGLE_SPREADSHEET_ID is defined.

@auryn-macmillan , VUE_APP_GOOGLE_SPREADSHEET_ID, seems to be a new feature introduced post v0.6.0, please confirm if this feature should be retained.

yuetloo:
Changes:
- If the VUE_APP_GOOGLE_SPREADSHEET_ID environment variable is set, users will be asked to provide email on the recipient registration flow.  This fixes #516 

- Added recipient registration form validation
- Removed universal recipient registry, use existing optimistic recipient registry instead
- Moved the transaction submit button on the Metadata form submission widget to the Form Navigation Panel.


auryn-macmillan:
image
I wonder if we should either remove the categories or allow the owner to define them somehow? These were specific to the ETHStaker round and might not make sense in other contexts.

auryn-macmillan (approved):
lgtm

yuetloo:
I could remove the category for this PR.

 Allowing owner to define categories is a good idea. In fact, may be have a separate PR to allow owner to define any fields?

auryn-macmillan:
> I could remove the category for this PR.

I think this is fine.

> Allowing owner to define categories is a good idea. In fact, may be have a separate PR to allow owner to define any fields?

Yeah, doing it as a separate feature and PR makes sense.

PR #540, opened 2022-08-10
Upgrade to brightId v6 api

yuetloo:
This PR..
- use BrightId v6 api
- use blind signatures

@auryn-macmillan , this PR has breaking change. We should tag the `develop` branch before merging.

To avoid re-deploying the funding round factory, I didn't make any change to the `deployNewRound()` function, instead, I made changes to the deployRound scripts to set the registration period after starting new round.


docs/brightid.md
```diff
@@ -18,8 +18,12 @@ USER_REGISTRY_TYPE=brightid
 
 Available envs:
 
-- Testing: `CLRFundTest`
-- Production: `clr.fund`
+| Network/Env | Context | Sponsor Contract |
+| ----------- | ------- | ---------------- |
+| goerli | clrfund-goerli | 0xF045234A776C87060DEEc5689056455A24a59c08 |
+| xdai | clrfund-gnosis-chain ||
```
auryn-macmillan, line 24: Shall we go ahead and deploy sponsor contracts to each of these networks?

auryn-macmillan (commented):
Looks good to me!

auryn-macmillan (approved):

docs/brightid.md
yuetloo: The sponsor contract has been deployed to goerli. As for xdai, since it's production, shall we wait until audit is done and deploy it when starting the next round?

auryn-macmillan: I think we're pretty safe on this particular contract, so I'd be in favor of just going ahead and deploying. The only reason I'd want to hold off on this one is if we predict it changing again.

PR #544, opened 2022-08-18
Upload ipfs file using pinata

yuetloo:
This PR fixes the image upload error when adding a project.

Infura started to enforce api key for using their IPFS upload api. Since the existing ipfs-mini library does not have the option to pass the api key, I changed the code to use piñata which doesn't require credit card to obtain the api key for development.


yuetloo:
close as code is in the brightid-v6 branch

PR #545, opened 2022-08-18
Fix error mapping simple user registry

yuetloo:
The fundingRoundFactory mapping can fail when mapping simple user registry because it's missing baseDeposit and challengePeriod.


yuetloo:
close as this is in the brightid-v6 branch as well.

PR #553, opened 2022-09-09
Quadratic funding calculation to include alpha

yuetloo:
This change implements the formula as described in https://arxiv.org/pdf/1809.06421.pdf, page 17, which uses alpha as an adjustment to avoid exceeding the funding budget.


contracts/contracts/FundingRound.sol
```diff
@@ -263,7 +315,12 @@ contract FundingRound is Ownable, MACISharedObjs, SignUpGatekeeper, InitialVoice
     // Total amount of spent voice credits is the size of the pool of direct rewards.
     // Everything else, including unspent voice credits and downscaling error,
     // is considered a part of the matching pool
-    matchingPoolSize = nativeToken.balanceOf(address(this)) - totalSpent * voiceCreditFactor;
+    uint256 budget = nativeToken.balanceOf(address(this));
+    matchingPoolSize = budget - totalSpent * voiceCreditFactor;
+
+    // TODO: need to check/verify totalVotesSquares
+    alpha = calcAlpha(budget, totalVotesSquares, _totalSpent);
```
auryn-macmillan: Is there a reason to pass in `_totalSpent` rather than `totalSpent` ?

auryn-macmillan (approved):

contracts/contracts/FundingRound.sol
```diff
@@ -456,8 +456,8 @@ contract FundingRound is Ownable, MACISharedObjs, SignUpGatekeeper, InitialVoice
     uint256 _tallyResultSalt
   )
     external
-    onlyOwner
   {
+    require(msg.sender == coordinator, 'FundingRound: Sender is not the coordinator');
```
auryn-macmillan: Since this check is being used in multiple functions, perhaps we should make it a modifier like in `FundingRoundFactory.sol` ?

```solidity
  modifier onlyCoordinator() {
    require(msg.sender == coordinator, 'FundingRound: Sender is not the coordinator');
    _;
  }
  ```

auryn-macmillan (approved):
lgtm

contracts/contracts/FundingRound.sol
yuetloo: Not really any reason, changed it to `totalSpent`. I don't think there's much difference as both should be available on the stack.

auryn-macmillan (approved):
lgtm

auryn-macmillan (approved):

PR #556, opened 2022-09-13
add etherscan verify plugin

yuetloo:
This PR is to add the ability to verify contracts from the command line. Also added a script to verify the MACI contract as its constructor arguments are complicated to handcraft.


auryn-macmillan (approved):
looks good!

PR #563, opened 2022-09-22
Use pinata api key and secret api key to authenticate

yuetloo:
This PR adds an alternative way to authenticate with Pinata because the JWT authentication cause issues in deploying netlify functions due to environment variable values over the 4k limit.


auryn-macmillan (approved):
lgtm

PR #577, opened 2022-10-09
Fix BrightId mobile link

yuetloo:
Use the BrightId universal mobile link instead of the schema link because clrfund incorrectly redirects schema link as internal link instead of external link and as a result cannot open the BrightId mobile app.

Also changed the BrightId status polling logic to only poll on timeout instead of on interval to avoid overloading the BrightId node potentially causing 503 service unavailable error.


auryn-macmillan (approved):
Looks good to me

PR #581, opened 2022-10-10
Increase recipient query row count limit

yuetloo:


vue-app/src/graphql/API.ts
```diff
@@ -2813,7 +2842,7 @@ export const GetRecipientDonationsDocument = gql`
     `;
 export const GetRecipientsDocument = gql`
     query GetRecipients($registryAddress: String!) {
-  recipients(where: {recipientRegistry: $registryAddress}) {
+  recipients(first: 300, where: {recipientRegistry: $registryAddress}) {
```
auryn-macmillan, line 2845: Is there a reason to limit it to 300? If we remove the "first" filter, out defaults to 1000, no?

auryn-macmillan (approved):

vue-app/src/graphql/API.ts
yuetloo: Yeah, performance concern. According to the [doc](https://thegraph.com/docs/en/querying/querying-best-practices/#performance-tips), the default is 100 (without specifying the `first` keyword. The max is 1000 when you specify using the `first` keyword. I tested and confirmed the default 100 rows per query.

We should do paging as a long time solution: https://github.com/graphprotocol/graph-client/blob/main/packages/auto-pagination/README.md

PR #584, opened 2022-11-07
Use Unitap to sponsor user as a temporary workaround

yuetloo:
Users occasionally get 'timeout waiting for sponsorship' from BrightID.

This PR is to temporary use Unitap to workaround the sponsorship issue…

A longer term solution is to switch to the sponsorship solution that does not use event logging on smart contract.


auryn-macmillan (approved):
lgtm! 🎉

PR #585, opened 2022-11-08
Update text in the matching pool contribution modal

yuetloo:
This is to give users more information about where their funds are going.


auryn-macmillan (approved):
lgtm

PR #588, opened 2022-11-10
Cart not opening when DAI balance < 0.0000001 DAI

yuetloo:
This is a temporary fix, a more permanent fix should refactor the formatAmount function to break down the function and to avoid potentially converting big numbers into javascript number which is smaller than big number.

The bug being fixed here is to avoid ethers library throwing when it detected unsafe number passed to the commify function.


auryn-macmillan (approved):
Lgtm

PR #591, opened 2022-11-12
Fix error generating proofs for test circuits

yuetloo:
When the batch size was incorrectly set to the default of 8, zk proof generation will fail with `incorrect circuit`


auryn-macmillan (approved):
lgtm

PR #593, opened 2022-11-17
Stop polling brightID node for verification status

yuetloo:
Changed the flow of linking with BrightID to mimic Unitap so that we don't overload brightID node by polling it 5 seconds.


auryn-macmillan (approved):
lgtm

PR #612, opened 2022-12-15
Only translate category if value is available and only process markdown if text is available

yuetloo:
On the clrfund canonical site, the category is not available on the project profile causing a lot of errors logged in the console log.
Projects also missing text to be rendered with markdown causing lots of error on the console log.  Do not call markdown, if text is missing.


auryn-macmillan (approved):
looks good

PR #613, opened 2022-12-20
Static site after finalize

### LINKED ISSUE #537: Optimize round information retrieval
Currently, every minute, through the LOAD_ROUND_INFO action, the web app makes many calls to a node to get round information from these lines here: https://github.com/clrfund/monorepo/blob/develop/vue-app/src/api/round.ts#L91-L97

All of those calls can be reduced into a single subgraph call to retrieve the information from the fundingRounds subgraph entity.

Note: make sure the subgraph has all the information first. I have seen scenario where the subgraph entity is not updated until later because there's missing events to trigger the update.

### LINKED ISSUE #603: Generate static site for historical rounds
The idea is generate JSON file with all the information needed to display the leaderboard and project information for historical rounds. For historical rounds, the clrfund app will pull information from the generated JSON file instead of from the subgraph or contracts.  This solution is to address evolving contract interface.

There will be no wallet connection for historical rounds.

yuetloo:
Show the leaderboard page if a round is finalized and the round data is exported in JSON file.  Location of the file is set in the environment variable: VUE_APP_STATIC_ROUNDS_BASE_URL

If a round is finalized successfully, the leaderboard page will be accessible.
If a round is canceled, the existing project list page will be accessible.


auryn-macmillan (approved):
lgtm

PR #615, opened 2023-01-03
Remove Gundb and generate MACI key from wallet signature

### LINKED ISSUE #500: Replace GunDB with something more web3 native
GunDB has been a good temporary solution while we wait for something more web3 native to mature. However, it is not particularly decentralized and has frequently caused issues like #420 when our GunDB node is down or the app is misconfigured.

After speaking with @corpetty last week, I'm convinced there are some web3 native options that we could use right now, which would be both more reliable and less trustful.

There are two options I'd like to explore:
1. **[Waku](https://waku.org/) / [Waku Connect](https://wakuconnect.dev/):** is a decentralized messaging protocol (the successor to Whisper, originally part of the Ethereum vision).
> Waku has its origins in the Whisper protocol, but optimized for scalability and better usability
> Waku removes centralized third parties from our messaging, enabling private, secure, censorship-free communication.
> Waku is designed for generalized messaging, enabling both human-to-human or machine-to-machine communication.

One nice quality of Waku, according to what I understand from @corpetty, is that messages persist for roughly 30 days. This is the perfect amount of time for them to be available to be used during a round, and then vanish into the ether once the round is over and they won't be required again.

2. **Deterministic key derivation:** using a [process similar to loopring](https://medium.loopring.io/looprings-new-approach-to-generating-layer-2-account-keys-4a16cc334906), we could deterministically generate the user's MACI keys based on their authenticated address, the address of the round, and some nonce/salt. This way, rather than storing the keys somewhere, we just recreate them locally on-the-fly.

yuetloo:


daodesigner:
🙋 How is the data that gets signed/hashed to get the key generated?

yuetloo:
> 🙋 How is the data that gets signed/hashed to get the key generated?

EIP-191 compliant signatures are generated for the following message:
https://github.com/clrfund/monorepo/blob/deterministic-key/vue-app/src/api/user.ts#L10

The signature is sha256 hashed and concat with a key index (for key changes) and sha256 hashed again.
https://github.com/clrfund/monorepo/blob/deterministic-key/maci-utils/src/maci.ts#L90

The private key is generated using the hash from above using the `genPrivKey()` which is modified from MACI's `genRandomBabyJubValue()`:
https://github.com/clrfund/monorepo/blob/deterministic-key/maci-utils/src/maci.ts#L53

MACI's `genRandomBabyJubValue()`
https://github.com/privacy-scaling-explorations/maci/blob/0a39b6728a3060f37bda14436e2c4c1439c39b2e/crypto/ts/index.ts#L138

contracts/tasks/changeKey.ts
```diff
@@ -0,0 +1,197 @@
...
+        'ERC20',
+        nativeTokenAddress
+      )
+      const tokenDecimals = await nativeToken.decimals()
+      console.log('Token decimal', tokenDecimals.toString())
+      console.log('voiceCreditFactor', voiceCreditFactor.toString())
+
+      let nonce = 1
+      const oldIndex = Number(keyNum || 1)
+      const newIndex = oldIndex + 1
```
auryn-macmillan: I wonder if we should use a salt here, rather than a counter?
This is probably harder to keep track of, but also less easily exploitable.

That said, perhaps it just puts us back where we started, since then the user must keep track of their salt :sweat_smile:

auryn-macmillan (approved):
lgtm

daodesigner:
> The private key is generated using the hash from above using the `genPrivKey()` which is modified from MACI's `genRandomBabyJubValue()`:
> 
> https://github.com/clrfund/monorepo/blob/deterministic-key/maci-utils/src/maci.ts#L53
> 
> 
> 
> MACI's `genRandomBabyJubValue()`
> 
> https://github.com/privacy-scaling-explorations/maci/blob/0a39b6728a3060f37bda14436e2c4c1439c39b2e/crypto/ts/index.ts#L138

Not sure how safe this is, signing a deterministic string is not a great source of entropy. Message signing attacks are pretty common now and it's easy to imagine a simple phishing attack. You could sign and hash a random string, but then might as well just have used the original genRandomBabyJubValue? I understand the goal here is to make key recovery easier, but this may be too easy 😊there may be a better way to achieve a similar outcome.
cc @chaosma

auryn-macmillan:
> I understand the goal here is to make key recovery easier, but this may be too easy blushthere may be a better way to achieve a similar outcome.

Nice, keen to hear suggestions here. Key loss has been far more common than compromise, so this is the primary concern currently. But if we can solve in a way that't better for both and doesn't introduce any additional infrastructure requirements, then this would be great!

Another option we had considered was using Waku to persist MACI keys as messages, leveraging the Waku nodes that Status runs. However, after talking with the VAC team, @yuetloo wasn't confident that we'd end up with better performance than what we got out of the gunDB instance we were running.

contracts/tasks/changeKey.ts
yuetloo: Yeah, the reason to use the deterministic incremental key is for the UI to be able to retrieve the contributor index after key change. 

Initially, I implemented to use a new wallet to generate the new key.  But, testing showed that I lost the contributor index after the key change.  The contributor index was returned from the contract after the signup call.

Key change events only emit the new public key, not the contributor index, which is needed to construct new messages for MACI.

https://github.com/privacy-scaling-explorations/maci/blob/master/contracts/sol/MACI.sol#L399
```
emit PublishMessage(_message, _encPubKey);
```

So, with the incremental index, I would go back to the first key to retrieve the contributor index.

The `changeKey` script is really only for testing purposes, not for production use.  I used it to test the flow of key change, retrieving data after key change and viewing on the UI.

yuetloo:
> Another option we had considered was using Waku to persist MACI keys as messages, leveraging the Waku nodes that Status runs. However, after talking with the VAC team, @yuetloo wasn't confident that we'd end up with better performance than what we got out of the gunDB instance we were running.

Yeah, the VAC team didn't recommend storing critical information on Waku because there's no guarantee of data availability.

yuetloo:
> Not sure how safe this is, signing a deterministic string is not a great source of entropy. Message signing attacks are pretty common now and it's easy to imagine a simple phishing attack. You could sign and hash a random string, but then might as well just have used the original genRandomBabyJubValue? I understand the goal here is to make key recovery easier, but this may be too easy 😊there may be a better way to achieve a similar outcome. cc @chaosma

I'd love to see a better implementation.  Ideally, I wish a wallet can encrypt messages for us. :)

I understand this solution isn't ideal, but, I think it's no worse than the previous implementation, which used to store the encrypted data on gundb.  And, the way to decrypt the data is to use the hash of the signature. Either way, we're still using the hash of the signature to decrypt some data. The difference now is the data is stored on chain instead of on gundb.

daodesigner:
> > I understand the goal here is to make key recovery easier, but this may be too easy blushthere may be a better way to achieve a similar outcome.
> 
> Nice, keen to hear suggestions here. Key loss has been far more common than compromise, so this is the primary concern currently. But if we can solve in a way that't better for both and doesn't introduce any additional infrastructure requirements, then this would be great!
> 
> Another option we had considered was using Waku to persist MACI keys as messages, leveraging the Waku nodes that Status runs. However, after talking with the VAC team, @yuetloo wasn't confident that we'd end up with better performance than what we got out of the gunDB instance we were running.

Will bring this up internally and see if we can't come up with some long term solution. Off the cuff, a simple improvement would be to sign something like a MACI mnemonic instead of a fixed string.

daodesigner (commented):
NACK; should implement key derivation from better source of entropy than fixed string as a safe interim to a longer term key recovery solution. Would not recommend this approach to key generation/recovery as a best practice.

auryn-macmillan:
> simple improvement would be to sign something like a MACI mnemonic instead of a fixed string

We'd then need users to backup that mnemonic, right?

I don't think this is ideal, because it adds onboarding friction with yet another seed for the user to secure. But at least backing up a mnemonic is a common pattern for web3 native users.

yuetloo:
> > > I understand the goal here is to make key recovery easier, but this may be too easy blushthere may be a better way to achieve a similar outcome.
> > 
> > 
> > Nice, keen to hear suggestions here. Key loss has been far more common than compromise, so this is the primary concern currently. But if we can solve in a way that't better for both and doesn't introduce any additional infrastructure requirements, then this would be great!
> > Another option we had considered was using Waku to persist MACI keys as messages, leveraging the Waku nodes that Status runs. However, after talking with the VAC team, @yuetloo wasn't confident that we'd end up with better performance than what we got out of the gunDB instance we were running.
> 
> Will bring this up internally and see if we can't come up with some long term solution. Off the cuff, a simple improvement would be to sign something like a MACI mnemonic instead of a fixed string.

What is a `MACI mnemonic`? Did you mean the MACI private key?
If it is acceptable to let users paste the key back to the clrfund app when they want to see what they contributed and reallocate funds, this is the simplest solution for now.

This also eliminates the signature request and makes it free for clrfund to support contract wallets.

The other option is if there's a MACI wallet that can encrypt/decrypt MACI messages as well as signing transactions.  But, also noticed metamask dropped support for their encrypt/decrypt methods https://medium.com/metamask/metamask-api-method-deprecation-2b0564a84686 for security reasons of leaking wallet private key.

If we want to use key management service to store MACI key outside of the app, then we need a service that can authenticate users by signature. Otherwise, we'll go back to the same gundb solution, which still needs to sign a fixed string for the password.

auryn-macmillan:
What are your thoughts on the MACI key being derived from the user signing a fixed string, plus some user generated salt (functionally equivalent to a pin number or passphrase)?

This way, the user only has to remember/backup this salt, is never pasting in a mnemonic (bit of an anti-pattern), and phishers would have to convince the user to both enter their salt and sign a message (rather than just inadvertently sign a message).

auryn-macmillan:
If we wanted to support contract wallets as well, then their MACI keys would need to be derived from something like a mnemonic that they can back up, since they can't sign a message in the same way.

yuetloo:
@daodesigner , I have reimplemented the MACI key generation using the public key from the [passkey](https://support.apple.com/en-us/HT213305) and would like to hear your opinion about this approach. 

According to the [webAuthn](https://www.w3.org/TR/webauthn-2/) standards, 3 components are involved:
1. Web application
2. Backend Server (with a database storing public keys for authentication)
3. Authenticator (that generates and stores the passkey)

Because Clrfund runs on the client side only (no backend server), we will not store the public key, instead, we will use ecrecover to recover the public key for generating the MACI key. Because of this, we currently only support the ECDSA p256 curve, which allows public key recovery from the signatures.

Note that only the website that owns the domain can prompt the users to sign a random message with the passkey they created on the website. So, only the website can get the signature to recover the public key.

Users can manage (rename/delete) the passkeys they created on the authenticator. Here's how to manage passkeys:
[Google passkey support](https://developers.google.com/identity/passkeys/supported-environments)
[Sign in with passkey on Apple](https://support.apple.com/en-ca/guide/iphone/iphf538ea8d0/ios#:~:text=On%20the%20sign%2Din%20screen%2C%20tap%20the%20account%20name%20field,to%20scan%20the%20QR%20code.)

Security wise, I think it's more secure in that users do not need to sign a static message with their wallets, and no need to remember passwords.

However, there are limitations:
1) can only use authenticators that support p256 curve, which should be mostly supported as this is one of defaults curve
2) by default, chrome stores the passkey on the chrome profile (tested on MacOS only), so, users will only be able to access their votes from the same browser. Apple users can select to store the passkey on the keychain by selecting "other device" when creating the passkey. This will allow vote access from iPhone if the keychain is synced on iCloud.

Here's a test site if you would like to see how the UX flow:
https://clrfund-testnet.netlify.app/#/

How to get the test DAI used by the app:
https://github.com/clrfund/monorepo/blob/deterministic-key/docs/testing.md

daodesigner:
Very cool, definitely looks way better off the bat, (though ideally you could use the private key) and will look through the changes/try it out tonight 😊. Great work!

yuetloo:
> Very cool, definitely looks way better off the bat, (though ideally you could use the private key) and will look through the changes/try it out tonight 😊. Great work!

Thank you, looking forward to your feedback

vue-app/src/utils/accounts.ts
```diff
@@ -43,3 +57,105 @@ export function renderAddressOrHash(
...
+  fundingRoundAddress: string
+): Promise<string> {
+  const credential = new Credential(address)
+
+  const key = await getEncryptionKey(address, fundingRoundAddress)
+
+  if (key) {
+    return key
+  } else {
+    const key = await credential.create()
```
daodesigner, line 93: is this the public key or the private key?

vue-app/src/utils/accounts.ts
```diff
@@ -43,3 +57,105 @@ export function renderAddressOrHash(
...
+    address,
+    getCartStorageKey(fundingRoundAddress)
+  )
+
+  if (!hasVoted && !uncommitedCart) {
+    return ''
+  }
+
+  const credential = new Credential(address)
+  const keys = await credential.get()
```
daodesigner, line 125: how does this get locked down? this just on the user's browser?

vue-app/src/api/passkey.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  async get(): Promise<string[]> {
+    const challenge = utils.randomBytes(32)
+    const publicKey = {
+      challenge,
+      timeout: TIMEOUT,
+    }
+
+    const credential = await navigator.credentials.get({
```
daodesigner, line 164: where does navigator come from?

vue-app/src/api/passkey.ts
```diff
@@ -0,0 +1,172 @@
...
+      publicKey,
+    })
+
+    return recoverPubKeyFromCredential(credential)
```
daodesigner, line 168: just return one of the private keys instead, which would be a better source of entropy

vue-app/src/api/passkey.ts
```diff
@@ -0,0 +1,172 @@
...
+ */
+export class Credential {
+  walletAddress: string
+
+  constructor(walletAddress: string) {
+    this.walletAddress = walletAddress
+  }
+
+  async create(): Promise<string> {
+    const challenge = utils.randomBytes(32)
```
daodesigner, line 116: user should save this somewhere, I assume this is the password?

daodesigner (commented):
any chance there's a diagram of all the interactions happening?

yuetloo:
> any chance there's a diagram of all the interactions happening?

Yes, added https://github.com/clrfund/monorepo/blob/deterministic-key/docs/passkey.mmd

vue-app/src/utils/accounts.ts
yuetloo: This is the public key of the passkey, we can't get the private key from the passkey

vue-app/src/utils/accounts.ts
yuetloo: This is between the user's browser and the authenticator.

vue-app/src/api/passkey.ts
yuetloo: navigator is the user's browser api

vue-app/src/api/passkey.ts
yuetloo: The private key is never exposed, what we get from the authenticator is only the signature of the data we asked it to sign. That's why we are recovering the public keys from the signature.

vue-app/src/api/passkey.ts
yuetloo: No, this is not a password, this is just a random piece of data that we ask the authenticator to sign. The standard suggests using random data to avoid replay attack, but, in our case, we never expose any data from the passkey other than using the public key to encrypt the votes on the browser, we could probably use a constant here. 

Obviously, this logic won't work if we ever exposed the public key or can trick the authenticator and user for the signature.  I've tried testing from different domains and passing different domain from the api, all tests didn't get the signature.

yuetloo:
@daodesigner , @auryn-macmillan

https://security.stackexchange.com/questions/269503/security-concerns-in-using-passkey-webauthn-to-encrypt-data/269515#269515

yuetloo:
changes merged

PR #632, opened 2023-02-15
Update brightid UX

yuetloo:
- Added options to sponsor user using the brightid api instead of via contract event logging.
- Moved the self sponsor page back to the verify flow.


auryn-macmillan (approved):
lgtm

PR #643, opened 2023-03-15
Fix missing claim button when multiple recipients share the same address

### LINKED ISSUE #604: Recipient Registry view showing recipients with invalid recipientId and address
Subgraph issues:
1) After a recipient claim funds, an invalid recipient record is recorded in the subgraph and this entry is shown as having an empty address in the recipient register view.

The invalid record was created with the recipient address as the recipientId instead of the actual recipientId.  Also noticed the same issue with the contribution mapping, e.g. invalid id was used

2) after removing a recipient, details of the recipient is cleared out in the subgraph, resulting empty project name displayed on the recipient registry view.

3) after a recipient is accepted, the requester address gets mapped to the account resolving the request, which is the owner of the registry. this means the original requester address is lost after the recipient is accepted

4) contributions entity is created incorrectly using the recipient address as a key to reference the recipient entity

yuetloo:
When multiple recipients use the same recipient address, after one recipient claimed the fund, the UI will not show the claim button for other recipients.

### DIFF
--- a/subgraph/generated/schema.ts
+++ b/subgraph/generated/schema.ts
@@ -1621,23 +1621,6 @@ export class Recipient extends Entity {
     }
   }
 
-  get donations(): Array<string> | null {
-    let value = this.get("donations");
-    if (value === null || value.kind == ValueKind.NULL) {
-      return null;
-    } else {
-      return value.toStringArray();
-    }
-  }
-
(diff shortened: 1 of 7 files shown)


yuetloo:
This will fix #604

auryn-macmillan (approved):
lgtm

PR #644, opened 2023-03-16
Recipient profile page

yuetloo:
Currently, after successfully adding the recipient, users might be redirected to an empty recipients page because the app does not wait for the data to be indexed in the subgraph.

This PR will wait for the data to be available in the subgraph and present the recipient profile page.


auryn-macmillan (approved):
lgtm

PR #645, opened 2023-03-16
Make dark theme the default theme

yuetloo:
This PR set the dark theme as the default theme because the app currently looks better in dark theme and when we do custom branding, this will make it easier to just customize the dark theme styles.


auryn-macmillan (approved):
lgtm!

PR #646, opened 2023-03-21
Fix BrightId sponsor error 68

yuetloo:
This PR is to ignore the BrightID sponsor error 68, The app generated id was sponsored before, so that users can link their wallet with their BrightID account on subsequent rounds.


auryn-macmillan (approved):
LGTM

PR #659, opened 2023-05-17
Hide the 'Get Started' button and show the 'Join round' callout

yuetloo:
On the landing page, when there's no scheduled round, we want to show the 'Join round' callout to attract projects to join the round to be scheduled in the future.


auryn-macmillan (approved):
lgtm

PR #664, opened 2023-06-01
Fix missing round error

yuetloo:
When a funding round has not been deployed, the prompt for user signature will give error because of missing round information.  Do not prompt for signature if there's no round.


auryn-macmillan (approved):
lgtm

PR #684, opened 2023-06-14
Fix 504 gateway timeout error in dev

yuetloo:
In dev mode, we occasionally get the 504 error while loading the maci-utils library, pre-bundling seems to solve the issue.


auryn-macmillan (approved):
Lgtm

PR #692, opened 2023-06-29
Fix incorrect formatted amounts when token decimals is not 18

yuetloo:
When the native token decimal is not 18 (i.e. USDC is 6), the formatted amounts showing on the round information page will be incorrect. For example, a contribution of 20 USDC will show up as 20T USDC.

Users also cannot contribute to the matching pool due to incorrect amount calculation.
This bug also show incorrect claim amount on the project page after the round is finalized.


crisgarner (approved):

auryn-macmillan (approved):
lgtm

PR #693, opened 2023-07-03
Reallocation being ignored

yuetloo:
Contribution/Reallocation can be ignored if a user had previously contributed to a previous round.  

The issue was caused by the incorrect subgraph schema for PublicKey. A MACI public key for different round can have different MACI 'stateIndex'.  Currently, the subgraph mapping logic for a public key only updates the public key for the signup event for the first round a user contributed to.  It skips updating the public key with latest `stateIndex` value for subsequent rounds that the user contributes to.


auryn-macmillan (approved):
lgtm

PR #697, opened 2023-07-10
Only hide project if the removal request is resolved

yuetloo:
This fix requires deployment of subgraph as there's mapping changes.


auryn-macmillan (approved):
lgtm

PR #704, opened 2023-07-28
Fix slow loading landing page

yuetloo:
Refactor the logic to filter out canceled rounds to avoid hitting the subgraph for a list of rounds causing the landing page slow to load.


auryn-macmillan (approved):
lgtm

PR #706, opened 2023-08-02
Add rerunable tally task

yuetloo:
Previous, if the tally script was stopped in the middle due to error, it cannot be resumed after the error was corrected. This task allows the tally task to resume by passing 2 additional parameters: --maci-logs and --maci-state-file


auryn-macmillan (approved):
This is really nice, Loo.
I like that it's a task now, and the ability to pass in the maci state file is a lifesaver for when things timeout unexpectedly.

PR #708, opened 2023-08-18
Make IPFS gateway url configurable from the environment file or as command line param

yuetloo:
The IPFS public gateway, ipfs.io, is not accessible. So, make sure we can configure the app or command line with our own gateway.


auryn-macmillan (approved):
lgtm

PR #709, opened 2023-08-19
Add snapshot user registry

yuetloo:
This user registry will add users who can be verified by the proof of token ownership at a snapshot block.


contracts/contracts/userRegistry/SnapshotUserRegistry.sol
```diff
@@ -0,0 +1,153 @@
...
+    minBalance = newMinBalance;
+
+    emit MinBalanceChanged(minBalance);
+  }
+
+  /**
+   * @dev Reject user
+    * @param _user user account address
+   */
+  function rejectUser(address _user) external onlyOwner {
```
auryn-macmillan: What are the purposes of `rejectUser` and `resetUser`?
How do we envision them being used in practice, @yuetloo?

contracts/tests/userRegistrySnapshot.ts
```diff
@@ -0,0 +1,144 @@
...
+        accountProofRlpBytes
+      )
+      await tx.wait()
+    } catch (err) {
+      console.log('error setting storage hash', err)
+      throw err
+    }
+  })
+
+  describe('Add user', function () {
```
auryn-macmillan, line 91: Would be good to add tests for users that currently meet the requirements, but did not at the snapshot block, and for users who do not currently meet the requirements, but did at the snapshot block.

contracts/contracts/userRegistry/SnapshotUserRegistry.sol
yuetloo: I was thinking to let the coordinator reject sybil accounts. But, you're right that in practice, it's too late as once the user is added, they would've contributed and in the round.

yuetloo:
Closing this PR as changes have been merged to https://github.com/clrfund/monorepo/pull/711

PR #710, opened 2023-08-21
Tally - do not fetch logs if tally hash was published

yuetloo:
Currently, the tally task only skips fetching MACI event logs if the MACI log file path is given.  During re-run, if the task already successfully completed tallying, and only need to re-run the part that upload tally result to the contract, it should not fetch MACI logs.


auryn-macmillan (approved):
lgtm

PR #712, opened 2023-09-01
Fix incorrect mapping of recipient reject status

yuetloo:
When a recipient application was rejected through `ChallengeRequest()`, the status on the registry view was incorrectly displayed at `Live`. It should be `Rejected`


auryn-macmillan (approved):
lgtm

PR #718, opened 2023-11-04
Make recipient id unique for different registries

yuetloo:
This fix is important for subgraph mapping as the recipient id is used as the primary key in the `Recipient` entity. Duplicate keys can be generated if the recipiients from different registries have the exact same metadata and the recipientIndex can get overwritten.


auryn-macmillan (commented):
lgtm

PR #719, opened 2023-11-08
Fix recipient registry subgraph mapping

yuetloo:
This PR ensures that when processing subgraph event, 'AddRecipient', the recipient registry entity exists.  The recipient registry entity could be missing in the case when new recipient registry is set in the funding round factory because the setRegistry event is missing in the contract.


auryn-macmillan (approved):
Lgtm

PR #727, opened 2024-03-02
added Zupass User Registry

crisgarner:


contracts/contracts/userRegistry/ZupassUserRegistry.sol
```diff
@@ -0,0 +1,75 @@
...
+    view
+    returns (bool)
+  {
+    return users[_user];
+  }
+
+  /**
+    * @dev Check if the semaphore Id is verified.
+    */
+  function isVerifiedSemaphoreId(uint256 _semaphoreId)
```
yuetloo, line 69: do we need to check isVerifiedSemaphoreId() before adding the user in addUser()?

contracts/contracts/userRegistry/ZupassUserRegistry.sol
```diff
@@ -0,0 +1,76 @@
...
+    * @dev Add verified unique user to the registry.
+    */
+  function addUser(address _user, uint256 _semaphoreId)
+    external
+    onlyOwner
+  {
+    require(_user != address(0), 'UserRegistry: User address is zero');
+    require(!users[_user], 'UserRegistry: User already verified');
+    require(!semaphoreIds[_semaphoreId], 'UserRegistry: Semaphore Id already registred' );
+    users[_user] = true;
```
yuetloo, line 33: typo: `registred` should be `registered`

yuetloo:
@crisgarner thank you for the PR.

contracts/contracts/userRegistry/ZupassUserRegistry.sol
```diff
@@ -0,0 +1,77 @@
...
+pragma solidity ^0.6.12;
+
+import '@openzeppelin/contracts/access/Ownable.sol';
+
+import './IUserRegistry.sol';
+
+/**
+ * @dev A simple user registry managed by a trusted entity.
+ */
+contract ZupassUserRegistry is Ownable, IUserRegistry {
```
samajammin, line 12: Nitpick but I'd say `SemaphoreUserRegistry` would be more fitting for what this registry does

yuetloo: > Nitpick but I'd say `SemaphoreUserRegistry` would be more fitting for what this registry does

I'm happy to change the name to `SemaphoreUserRegistry`. This just means that the registry can be used to integrate with any system that can generate a unique semaphore id, not just by Zupass.

PR #731, opened 2024-03-26
Add ClrFundDeployer subgraph mapping

yuetloo:
The ClrFundDeployer subgraph mapping will track new instances of ClrFund. This allows one subgraph to track all ClrFund instances created by the ClrFundDeployer.


auryn-macmillan (approved):
lgtm

PR #738, opened 2024-04-03
Refactor tally hardhat script

yuetloo:
Refactor the tally script to simplify rerun on error. Rerun no longer need additional arguments.  Also break the tally script into the following smaller scripts to simplify re-running the scripts.

- gen-proofs
- prove-on-chain
- publish-tally-results


auryn-macmillan (approved):
lgtm

PR #740, opened 2024-04-16
Refactor contract verification to get constructor arguments from contracts

yuetloo:
This PR allows contract verifications to get constructor arguments from contracts if the arguments are not found in the deployed-contracts.json file.  

When new rounds are deployed from a Gnosis Safe wallet, the deployed-contracts.json file will not be generated (as it's generated when deploying from the hardhat command line), users can still verify the contracts by passing the `--clrfund` flag.


auryn-macmillan (approved):
Lgtm

PR #741, opened 2024-04-17
Refactor modals and disable click on modal background

yuetloo:
This PR prevents users from clicking on buttons in the modal background and interrupting transactions.


auryn-macmillan (approved):
Lgtm

PR #742, opened 2024-04-19
Enable closed funding rounds to be displayed statically without subgraph query

yuetloo:
This PR will display round information from the rounds.json file without querying the subgraph.

This is enabled when the VITE_SUBGRAPH_URL env. variable is not set.  It also require the round to be exported using `export-round` hardhat command.


auryn-macmillan (approved):
lgtm

PR #755, opened 2024-05-22
Fix incorrect max contributor causing frontend to prematurely closing a round

yuetloo:
This PR fixed the incorrect max contributor calculated using TREE_ARITY = 2 instead of 5. Because the max contributor was incorrectly calculated as too small, the frontend closes the round to prevent more contribution.

The fix also removed the TREE_ARITY constant in the contract as the TREE_ARITY can change from version to version in MACI.

The changes in the subgraph also allows upgraded FundingRound and MACI contracts to use older recipient registry to help migration into the new version of MACI contracts.


auryn-macmillan (commented):
lgtm

PR #759, opened 2024-06-04
Fix reallocation and contract verification errors

yuetloo:
Fixes in this PR:
1. Fix the contract verification errors caused by missing constructor arguments for Tally and MessageProcessor.
2. Add verification of user registry and recipient registry added to the ClrFund contract
3. Add sepolia etherscan url for the round export task
4. Fix reallocation error due to missing key caused by incorrect subgraph id contruction


auryn-macmillan (approved):
lgtm

